PR #10, opened 2015-05-18
Implemented CLI tool

vso-tc:


core/src/main/java/io/github/tcdl/cli/CliTool.java
```diff
@@ -1,26 +1,122 @@
...
+                    return Arrays.asList(optionValue.split(","));
+                }
+            }
+        }
+
+        return null;
+    }
+
+    private static void subscribe(String topicName, CliMessageHandler handler) {
+        synchronized (registeredTopics) {
```
vso-tc, line 74: This code is invoked when we process response messages. Hence it can be called from different threads in parallel. To ensure consistency of the set we have to synchronize the block.

Any other suggestions?

PR #13, opened 2015-05-19
Get rid of Spring dependency

vso-tc:


amqp/pom.xml
```diff
@@ -1,42 +1,43 @@
 <?xml version="1.0"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
vso-tc: Reformatted this file according to our format

vso-tc:
Pom files were reformatted. That is why the diff is so big

core/pom.xml
```diff
@@ -30,10 +31,6 @@
             <artifactId>commons-io</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.springframework.amqp</groupId>
```
vso-tc, line 33: This is the main change

rkatsyuryna:
well done

PR #14, opened 2015-05-19
Refactoring of event emitter to use lambda callbacks instead of event…

rdrozdov-tc:
… adapters

### DIFF
--- a/core/src/main/java/io/github/tcdl/events/Event.java
+++ b/core/src/main/java/io/github/tcdl/events/Event.java
@@ -3,31 +3,30 @@ package io.github.tcdl.events;
 /**
  * Created by rdro on 4/23/2015.
  */
-public class Event {
+public enum Event {
+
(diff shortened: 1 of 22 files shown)


core/src/main/java/io/github/tcdl/events/Event.java
```diff
@@ -3,31 +3,30 @@
...
-public class Event {
+public enum Event {
+
+    PRODUCER_NEW_TOPIC_EVENT("newProducerOnTopic"),
+    PRODUCER_NEW_MESSAGE_EVENT("newProducedMessage"),
+    CONSUMER_NEW_TOPIC_EVENT("newConsumerOnTopic"),
+    CONSUMER_REMOVED_TOPIC_EVENT("removedConsumerOnTopic"),
+    CONSUMER_NEW_MESSAGE_EVENT("newConsumedMessage"),
+
+    MESSAGE_EVENT("message"),
```
vso-tc, line 14: It would be great to have comments here what each of the events mean

PR #22, opened 2015-05-20
Added AmqpAdapterTest

vso-tc:


core/src/main/java/io/github/tcdl/adapters/AmqpAdapter.java
```diff
@@ -43,14 +43,14 @@ public void publish(String jsonMessage) throws ChannelException {
...
-    public void subscribe(RawMessageHandler msgHandler) {       
+    public void subscribe(RawMessageHandler msgHandler) {
         String groupId = configuration.getAmqpBrokerConf().getGroupId();
         boolean durable = configuration.getAmqpBrokerConf().isDurable();
 
         String queueName = generateQueueName(topic, groupId, durable);
 
         try {
-            channel.queueDeclare(queueName, durable /* durable */, false /* exclusive */, true /*auto-delete */, null);
+            channel.queueDeclare(queueName, durable /* durable */, false /* exclusive */, !durable /*auto-delete */, null);
```
vso-tc, line 53: This is an actual bug that was spotted by the new unit test

core/src/main/java/io/github/tcdl/adapters/AmqpConnectionManager.java
```diff
@@ -7,7 +7,9 @@
 import java.io.IOException;
 
 public class AmqpConnectionManager {
-    private static AmqpConnectionManager instance = new AmqpConnectionManager();
+    private static class LazyHolder {
```
vso-tc, line 10: Lazy holder idiom for lazy initialization of singletons. Otherwise if this class is initialized eagerly then connection attempt is made before we have a chance to set some expectations via Powermock

PR #23, opened 2015-05-20
remove MockAdapter from unit tests

rkatsyuryna:


core/src/itest/java/io/github/tcdl/RequesterIT.java
```diff
@@ -34,12 +34,12 @@
 
     private MsbMessageOptions messageOptions;
     private MsbConfigurations msbConf;
+    private String topic = "test:requester";
```
vso-tc: Probably it should be a constant

core/src/main/java/io/github/tcdl/adapters/MockAdapter.java
```diff
@@ -1,60 +1,73 @@
...
+        } catch (JsonConversionException e) {
+            log.error("Recieved message can not be parsed");
+        }
+    }
+
+    @Override
+    public void subscribe(RawMessageHandler messageHandler) {
+    }
+
+    public void subscribeForTopic(RawMessageHandler messageHandler, String topic) {
```
vso-tc: Maybe it would be better to create separate adapter per topic as implemented for AmqpAdapter and properly implement subscribe method instead of adding this public method?

core/src/main/java/io/github/tcdl/Producer.java
```diff
@@ -1,58 +1,61 @@
-package io.github.tcdl;
```
vso-tc, line 1: Could you comment what actually changed in the class in such cases when due to difference in formatting the file looks completely different?

PR #27, opened 2015-05-21
Made CLI tool more robust in case of wrong response topic in the message

vso-tc:


core/src/main/java/io/github/tcdl/cli/CliMessageHandler.java
```diff
@@ -0,0 +1,57 @@
...
+            objectMapper.enable(SerializationFeature.INDENT_OUTPUT);
+        }
+
+        try {
+            JSONObject jsonMessageObject = new JSONObject(jsonMessage);
+
+            // Subscribe to additional topics found in the message
+            if (jsonMessageObject.has("topics")
+                    && jsonMessageObject.getJSONObject("topics").has("response")
+                    && jsonMessageObject.getJSONObject("topics").get("response") instanceof String
```
vso-tc, line 38: Bug 1. If response topic is not a string then the tool crashed

core/src/main/java/io/github/tcdl/cli/CliMessageHandler.java
```diff
@@ -0,0 +1,57 @@
...
+                    && follow.contains("response")) {
+
+                String responseTopicName = jsonMessageObject.getJSONObject("topics").getString("response");
+
+                try {
+                    subscriber.subscribe(responseTopicName, this);
+                } catch (Exception e) {
+                    // Just ignore the exception
```
vso-tc, line 46: Bug 2. If response topic does not exist then the tool crashed

core/src/main/java/io/github/tcdl/cli/CliMessageHandler.java
```diff
@@ -0,0 +1,57 @@
...
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import io.github.tcdl.adapters.Adapter;
+import org.json.JSONObject;
+
+import java.io.IOException;
+import java.util.List;
+
+class CliMessageHandler implements Adapter.RawMessageHandler {
```
vso-tc, line 12: Extracted the class to make it more testable

PR #28, opened 2015-05-21
Removed RabbitMQ related classes to separate amqp project

YuriySavchuk:
Please apply changes for creation separate project with AMQP adapter.

### DIFF
--- a/amqp/src/main/java/io/github/tcdl/adapters/amqp/AmqpAdapter.java
+++ b/amqp/src/main/java/io/github/tcdl/adapters/amqp/AmqpAdapter.java
@@ -1,13 +1,18 @@
-package io.github.tcdl.adapters;
+package io.github.tcdl.adapters.amqp;
 
 import com.rabbitmq.client.*;
 
+import io.github.tcdl.adapters.Adapter;
 import io.github.tcdl.config.MsbConfigurations;
 import io.github.tcdl.exception.ChannelException;
 
 import java.io.IOException;
 
 import org.apache.commons.lang3.Validate;
+
+/*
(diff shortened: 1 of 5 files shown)


amqp/src/main/java/io/github/tcdl/adapters/amqp/AmqpAdapter.java
```diff
@@ -75,7 +95,15 @@ public void unsubscribe() {
         }
     }
 
+    /**
+     * Represents the topic name for logging purpose
```
vso-tc, line 99: This is not entirely correct description. We generate topic name in this way to get unique topics for different microservices. That's the main point of this code.

core/src/main/java/io/github/tcdl/adapters/AdapterFactory.java
```diff
@@ -19,7 +22,17 @@ public static AdapterFactory getInstance() {
 
     public Adapter createAdapter(BrokerAdapter brokerName, String topic, MsbConfigurations msbConfig) {
         if (brokerName == BrokerAdapter.AMQP) {
-            return new AmqpAdapter(topic, msbConfig);
+            try {
+                Class<?> clazz = Class.forName("io.github.tcdl.adapters.amqp.AmqpAdapter");
+                Constructor<?> constructor = clazz.getConstructor(String.class, MsbConfigurations.class);
```
vso-tc, line 27: A lot of reflection here. Maybe adapters shouldn't be created via direct constructor invocation but via call to some factory? Then we could extract the interface of that factory and call it more conveniently. Also it will enforce the interface for adapter creation (in our case topic name and configuration instance).

PR #39, opened 2015-05-25
Responder utit test

rdrozdov-tc:


core/src/test/java/io/github/tcdl/ResponderServerTest.java
```diff
@@ -1,23 +1,50 @@
...
+
+        ResponderServer
+                .create(TestUtils.createSimpleConfig())
+                .use(middleware)
+                .listen();
+
+        mockStatic(CompletableFuture.class);
+        ArgumentCaptor<Supplier> middlewareCaptor = ArgumentCaptor.forClass(Supplier.class);
+
+        ChannelManager.getInstance().emit(Event.MESSAGE_EVENT, TestUtils.createMsbRequestMessageWithPayload());
```
vso-tc, line 41: Why do we need to invoke another class to tests this one? Can't we invoke some public method on ResponderServer directly?

PR #53, opened 2015-05-26
-Add test case for Responder

anstr:
-Refactor test case for Requester

### DIFF
--- a/core/src/test/java/io/github/tcdl/ResponderTest.java
+++ b/core/src/test/java/io/github/tcdl/ResponderTest.java
@@ -7,7 +7,6 @@ import io.github.tcdl.support.TestUtils;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import org.mockito.exceptions.verification.TooLittleActualInvocations;
 import org.powermock.api.mockito.PowerMockito;
 import org.powermock.core.classloader.annotations.PrepareForTest;
 import org.powermock.modules.junit4.PowerMockRunner;
@@ -50,7 +49,6 @@ public class ResponderTest {
 
     @Test
     public void testResponderNotNullOk() {
-        String topic = config.getNamespace();
         new Responder(config, message);
     }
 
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/github/tcdl/ResponderTest.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void testResponderNullConfigsThrowsException() {
+        new Responder(null, message);
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void testResponderNullMessageThrowsException() {
+        new Responder(config, null);
```
rkatsyuryna, line 47: looks like here we can simply mock MsbConfig

core/src/test/java/io/github/tcdl/ResponderTest.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Test
+    public void testResponderNotNullOk() {
+        String topic = config.getNamespace();
```
rkatsyuryna: and also here..

core/src/test/java/io/github/tcdl/ResponderTest.java
```diff
@@ -0,0 +1,94 @@
...
+    public void testResponderSendOk() throws Exception {
+        Responder spy = PowerMockito.spy(responder);
+
+        Payload payload = new Payload.PayloadBuilder().build();
+        spy.send(payload, null);
+
+        PowerMockito.verifyPrivate(spy, times(1)).invoke("send", anyObject(), anyObject());
+    }
+
+    @Test(expected = TooLittleActualInvocations.class)
```
rkatsyuryna: not sure we need this check. isn't testResponderSendAckOk() not verifying send() is called exactly once?

core/src/test/java/io/github/tcdl/ResponderTest.java
```diff
@@ -0,0 +1,94 @@
...
+    public void testResponderSendFailFailNotCorrectNumberOfInvokes() throws Exception {
+        Responder spy = PowerMockito.spy(responder);
+
+        Payload payload = new Payload.PayloadBuilder().build();
+        spy.send(payload, null);
+
+        PowerMockito.verifyPrivate(spy, times(2)).invoke("send", anyObject(), anyObject());
+    }
```
rkatsyuryna, line 72: not sure we need this check. isn't testResponderSendOk() not verifying send() is called exactly once?

core/src/test/java/io/github/tcdl/ResponderTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Test
+    public void testResponderSendOk() throws Exception {
+        Responder spy = PowerMockito.spy(responder);
+
+        Payload payload = new Payload.PayloadBuilder().build();
+        spy.send(payload, null);
+
+        PowerMockito.verifyPrivate(spy, times(1)).invoke("send", anyObject(), anyObject());
```
vso-tc: Is this a typo? Should be "sendMessage"?

core/src/test/java/io/github/tcdl/ResponderTest.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Test
+    public void testResponderNotNullOk() {
+        String topic = config.getNamespace();
+        new Responder(config, message);
+    }
+
+    @Test
+    public void testResponderSendAckOk() throws Exception {
```
vso-tc, line 56: What's the purpose of this test? I doubt that we need to test whether some private method of the SAME class is invoked. Also this approach is very fragile in case of  refactorings.

I assume that we need to check here that proper message is published via produce mock. Do we have such test?

core/src/test/java/io/github/tcdl/ResponderTest.java
vso-tc, line 72: +1

core/src/test/java/io/github/tcdl/ResponderTest.java
anstr: Agree)It is unnecessary)Will be removed)

core/src/test/java/io/github/tcdl/ResponderTest.java
anstr, line 72: Agree)It is unnecessary)Will be removed)

core/src/test/java/io/github/tcdl/ResponderTest.java
anstr: Yes,it's typo)

PR #62, opened 2015-05-27
Unit tests

anstr:


core/src/test/java/io/github/tcdl/ResponderTest.java
```diff
@@ -0,0 +1,103 @@
...
+    public void setUp() {
+        config = TestUtils.createSimpleConfig();
+
+        msbConf = TestUtils.createMsbConfigurations();
+
+        MessageFactory messageFactory = new MessageFactory(msbConf.getServiceDetails());
+        MessageFactory spyMessageFactory = spy(messageFactory);
+
+        MsbContext msbContext = TestUtils.createSimpleMsbContext();
+        MsbContext msbContextSpy = spy(msbContext);
```
vso-tc, line 46: Why do you use 2 instances of MsbContext? MsbContext is almost a pojo. Why can't you inject mockChannelManager directly?

core/src/test/java/io/github/tcdl/ResponderTest.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Test
+    public void testProducerWasCreatedForProperTopic() {
+        ArgumentCaptor<String> argument = ArgumentCaptor.forClass(String.class);
+        responder.send(payload, null);
+
+        verify(mockChannelManager).findOrCreateProducer(argument.capture());
+
+        assertEquals(originalMessage.getTopics().getResponse(), argument.getValue());
```
vso-tc, line 76: Seems like you don't need argument captor to make this verification

core/src/test/java/io/github/tcdl/ResponderTest.java
```diff
@@ -0,0 +1,103 @@
...
+    @Test
+    public void testProducerPublishMethodInvoked() {
+        responder.send(payload, null);
+
+        verify(mockProducer, times(1)).publish(anyObject(), anyObject());
+    }
+
+    @Test
+    public void testProducerPublishUseCorrectPayload() {
+        ArgumentCaptor<Message> argument = ArgumentCaptor.forClass(Message.class);
```
vso-tc, line 88: Again, seems like you don't need argument captor here.

core/src/test/java/io/github/tcdl/ResponderTest.java
```diff
@@ -0,0 +1,103 @@
...
+        verify(mockProducer).publish(argument.capture(), anyObject());
+
+        assertEquals(payload, argument.getValue().getPayload());
+    }
+
+    @Test
+    public void testAckBuilderContainsCorrectProperties() {
+        responder.sendAck(200, 2, null);
+
+        assertEquals(acknowledgeBuilder.getResponsesRemaining().intValue(), 2);
```
vso-tc, line 100: Would it be better to check correctness of the message via checking of ack message in the previous method?

PR #79, opened 2015-05-29
Travis deploy snapshot

anstr:


vso-tc:
Could you squash all the changes in a single commit? 13 commits seem to be too many.

PR #81, opened 2015-06-02
Event emitter to consumer

rdrozdov-tc:


core/src/itest/java/io/github/tcdl/RequesterResponderIT.java
```diff
@@ -84,6 +84,7 @@ public void testResponderAnswerWithAckRequesterRecieveAck() throws Exception {
                 }))
                 .listen();
 
+        //Thread.sleep(60000);
```
vso-tc, line 87: Why do we need this? If there's a reason please put a comment

core/src/main/java/io/github/tcdl/Collector.java
```diff
@@ -124,8 +122,11 @@ protected void cancel() {
 
     protected void end() {
         LOG.debug("End");
+        Consumer consumer = channelManager.findConsumer(this.topic);
```
vso-tc, line 125: Looks like a hack

core/src/main/java/io/github/tcdl/ChannelManager.java
```diff
@@ -68,6 +63,10 @@ public Consumer findOrCreateConsumer(final String topic) {
         return consumer;
     }
 
+    public Consumer findConsumer(final String topic) {
```
vso-tc, line 66: Why do we need this. Isn't findOrCreateConsumer enough?

core/src/main/java/io/github/tcdl/ChannelManager.java
```diff
@@ -79,7 +78,7 @@ public void removeConsumer(String topic) {
         channelMonitorAgent.consumerTopicRemoved(topic);
     }
 
-    private Producer createProducer(String topic) {
+    public Producer createProducer(String topic) {
```
vso-tc, line 81: Again, doesn't it break encapsulation?

core/src/main/java/io/github/tcdl/ChannelManager.java
```diff
@@ -87,32 +86,14 @@ private Producer createProducer(String topic) {
         return new Producer(adapter, topic, handler);
     }
 
-    private Consumer createConsumer(String topic) {
+    public Consumer createConsumer(String topic) {
```
vso-tc, line 89: Again, doesn't it break encapsulation?

core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -2,16 +2,22 @@
 
 import io.github.tcdl.adapters.Adapter;
 import io.github.tcdl.config.MsbConfigurations;
-import io.github.tcdl.events.TwoArgsEventHandler;
+import io.github.tcdl.events.*;
```
vso-tc, line 5: Please fix your IDE settings

core/src/main/java/io/github/tcdl/ChannelManager.java
```diff
@@ -87,32 +86,14 @@ private Producer createProducer(String topic) {
...
-            }
-
-            if (isMessageExpired(message))
-                return;
-            channelMonitorAgent.consumerMessageReceived(topic);
-            emit(MESSAGE_EVENT, message);
-        };
-
-        return new Consumer(adapter, topic, handler, msbConfig);
+        return new Consumer(adapter, topic, new EventEmitterImpl(), msbConfig, channelMonitorAgent);
```
vso-tc, line 94: Why do we need this event emitter impl if we don't subscribe to it?

core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -68,4 +57,58 @@ public void end() {
...
+
+        if (error != null) {
+            eventEmitter.emit(ERROR_EVENT, error);
+            return;
+        }
+
+        if (isMessageExpired(message))
+            return;
+
+        if (channelMonitorAgent != null) {
```
vso-tc, line 101: This is a redundant check. Channel manager is never null in Channel manager so you can just put validation statement in constructor

core/src/main/java/io/github/tcdl/ResponderServer.java
```diff
@@ -42,32 +39,20 @@ public ResponderServer use(Middleware... middleware) {
...
+            onResponder(responder);
         });
 
-        channelManager.findOrCreateConsumer(topic);
-
         return this;
     }
 
-    private SingleArgEventHandler<Responder> onResponder = (Responder responder) -> {
+    protected void onResponder (Responder responder) {
```
vso-tc, line 55: Does it really need to be protected? Isn't package-private enough?

PR #84, opened 2015-06-03
Java time migration

vso-tc:


core/pom.xml
```diff
@@ -56,6 +56,10 @@
             <artifactId>json</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
```
vso-tc, line 59: This jackson extension enables support for java.time datatypes

core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -106,9 +115,15 @@ protected void handleRawMessage(String jsonMessage) {
     }
 
     private boolean isMessageExpired(Message message) {
-        return message.getMeta() != null
-                && message.getMeta().getTtl() != null
-                && DateUtils.addMilliseconds(message.getMeta().getCreatedAt(), message.getMeta().getTtl()).after(
-                new Date());
+        MetaMessage meta = message.getMeta();
```
vso-tc, line 118: Fixed actual bug in this method

core/src/main/java/io/github/tcdl/support/Utils.java
```diff
@@ -57,11 +57,11 @@ public static Integer getPid() {
...
-        ObjectMapper objectMapper =  new ObjectMapper().setDateFormat(JSON_DATE_FORMAT);
-        objectMapper.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
-        objectMapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
-
-        return objectMapper;
+        return new ObjectMapper()
+                .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
+                .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
+                .setSerializationInclusion(JsonInclude.Include.NON_NULL)
+                .registerModule(new JSR310Module());
```
vso-tc, line 64: JSR-310 is specification for the new Java date-time API

core/src/test/java/io/github/tcdl/messages/MessageFactoryTest.java
```diff
@@ -131,8 +134,7 @@ public void testCreateMessageBuilderMetaFromMsgOptions() throws Exception {
 
         Message message = mesageBuilder.build();
         assertEquals("Message ttl is not set correctly", ttl, message.getMeta().getTtl());
-        assertEquals("Message create_at is not equals now", DateUtils.truncate(new Date(), Calendar.SECOND),
-                DateUtils.truncate(message.getMeta().getCreatedAt(), Calendar.SECOND));
+        assertEquals("Message create_at is not equals now", FIXED_CLOCK_INSTANT, message.getMeta().getCreatedAt());
```
vso-tc, line 137: This is a place were the new approach really shines. Instead of non-reliable truncating that we had before we can check for exact equality.

PR #105, opened 2015-06-09
Consumer subscriber inroduced

rdrozdov-tc:


core/src/main/java/io/github/tcdl/ChannelManager.java
```diff
@@ -50,9 +50,15 @@ public synchronized Producer findOrCreateProducer(final String topic) {
         return producer;
     }
 
-    public synchronized Consumer findOrCreateConsumer(final String topic) {
+    public synchronized void subscribe(final String topic, Consumer.Subscriber subscriber) {
+        Consumer consumer = findOrCreateConsumer(topic);
+        consumer.subscribe(subscriber);
+    }
+
+    protected Consumer findOrCreateConsumer(final String topic) {
```
vso-tc: Why "protected"? Isn't package-private enough?

core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -31,7 +30,7 @@
     private ChannelMonitorAgent channelMonitorAgent;
     private Clock clock;
 
-    private List<Subscriber> subscribers;
+    private ConcurrentLinkedQueue<Subscriber> subscribers;
```
vso-tc, line 33: Do we really need to have concurrent collection here? Seems like methods that modify it are called from other synchronized methods. This double synchronization seems to be an overkill.

core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -42,25 +41,22 @@ public Consumer(Adapter rawAdapter, String topic, MsbConfigurations msbConfig, C
...
     }
 
-    public synchronized Consumer subscribe(Callback<Message> messageHandler, Callback<Exception> errorHandler) {
-        Validate.notNull(messageHandler, "the 'messageHandler' must not be null");
-
-        subscribers.add(new Subscriber(messageHandler, errorHandler));
-        rawAdapter.subscribe(this::handleRawMessage);
-
-        return this;
+    public boolean unsubscribe(Subscriber subscriber) {
```
vso-tc: Javadoc and unit test is required here to describe what return value means.

core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -91,21 +87,8 @@ protected void handleRawMessage(String jsonMessage) {
 
         synchronized (this) {
```
vso-tc: Can you explain why do we need this synchronized block?

core/src/test/java/io/github/tcdl/ChannelManagerTest.java
```diff
@@ -6,28 +6,29 @@
...
 
 /**
  * @author rdro
  * @since 4/24/2015
  */
+@RunWith(MockitoJUnitRunner.class)
 public class ChannelManagerTest {
 
     private ChannelManager channelManager;
     private ChannelMonitorAgent mockChannelMonitorAgent;
```
vso-tc: Should we put @Mock on ChannelMonitorAgent?

core/src/test/java/io/github/tcdl/CollectorTest.java
```diff
@@ -84,35 +74,35 @@ public void testCreateCollectorNullMsgOptions() {
     @Test
     public void testGetWaitForResponsesConfigsNoWaitForResponsesReturnFalse() {
         when(messageOptionsMock.getWaitForResponses()).thenReturn(null);
-        Collector collector = new Collector(messageOptionsMock, msbContext, eventHandlers);
+        Collector collector = spy(new Collector(messageOptionsMock, msbContext, eventHandlers));
```
vso-tc: Why do we need spy here and all the following methods?

core/src/test/java/io/github/tcdl/ConsumerTest.java
```diff
@@ -71,28 +69,31 @@ public void testCreateConsumerNullMsbConf() {
     @Test
     public void testConsumeFromTopicValidateThrowException() {
         MsbConfigurations msbConf = TestUtils.createMsbConfigurations();
-        Consumer consumer = new Consumer(adapterMock, TOPIC, msbConf, clock, channelMonitorAgentMock).subscribe(messageHandler, errorHandler);
+        Consumer consumer = new Consumer(adapterMock, TOPIC, msbConf, clock, channelMonitorAgentMock);
+        consumer.subscribe(subscriberMock);
 
         consumer.handleRawMessage("{\"body\":\"fake message\"}");
-        verify(errorHandler).call(isA(JsonSchemaValidationException.class));
+        verify(subscriberMock).handleMessage(any(), isA(JsonSchemaValidationException.class));
```
vso-tc, line 80: Why any()? Can't we have more strict validation?

PR #111, opened 2015-06-11
added a test for ConfigurationUtil

rdrozdov-tc:


core/src/test/java/io/github/tcdl/configuration/ConfigurationUtilTest.java
```diff
@@ -0,0 +1,158 @@
...
+        String param = "parameter.boolean";
+        Boolean paramValue = true;
+        Boolean fallback = true;
+
+        when(config.hasPath(param)).thenReturn(true);
+        when(config.getBoolean(param)).thenReturn(paramValue);
+
+        Boolean value = ConfigurationUtil.getBoolean(config, param, fallback);
+
+        assertEquals(paramValue, value);
```
rkatsyuryna: I would use assertTrue here. but actually i would like to see 3 cases here:

1.
when(config.hasPath(param)).thenReturn(true);
Boolean value = ConfigurationUtil.getBoolean(config, param, fallback);
verify(config).getBoolean(eq(param))
1. when(config.hasPath(param)).thenReturn(false);
   Boolean value = ConfigurationUtil.getBoolean(config, param, fallback);
   verify(config, never()).getBoolean(eq(param))
2.  when(config.hasPath(param)).thenReturn(true);
   Boolean value = ConfigurationUtil.getBoolean(config, param, fallback);  
   verify(config, never()).getBoolean(eq(param))
   assertEquals(fallback, value);

Or even combine 2 and 3 into one case

core/src/test/java/io/github/tcdl/configuration/ConfigurationUtilTest.java
```diff
@@ -0,0 +1,158 @@
...
+    @Test
+    public void testGetOptionalBooleanNotExists() {
+        String param = "parameter.boolean";
+
+        when(config.hasPath(param)).thenReturn(false);
+
+        Optional<Boolean> value = ConfigurationUtil.getOptionalBoolean(config, param);
+
+        assertFalse(value.isPresent());
+    }
```
rkatsyuryna: also would like to test interaction here. See comment above

PR #116, opened 2015-06-11
Amqp adapter tests

vso-tc:


rkatsyuryna:
plz create 1 commit for this pull request

PR #119, opened 2015-06-11
improved performance of json schema validation

rdrozdov-tc:


core/src/test/java/io/github/tcdl/support/JsonValidatorTest.java
```diff
@@ -0,0 +1,83 @@
...
+        try {
+            validator.validate(jsonMessage, messageSchema);
+        } catch(Exception e){
+            fail("Validation should not fail");
+        }
+    }
+
+    @Test
+    public void testCachingJsonSchema() throws Exception {
+        String jsonMessage = Utils.toJson(TestUtils.createMsbRequestMessageNoPayload());
```
vso-tc, line 51: Let's make this test more robust

PR #128, opened 2015-06-15
fix logic with ack scheduling, exception handling in case rawAdapter …

rkatsyuryna:
…throw the one, fix unnecessary collector creation


core/src/main/java/io/github/tcdl/Collector.java
```diff
@@ -78,11 +78,7 @@ public Collector(MsbMessageOptions messageOptions, MsbContext msbContext, EventH
         }
     }
 
-    public boolean isWaitForResponses() {
-        return waitForResponses != 0;
-    }
-
-    public boolean isAwaitingAcks() {
+    boolean isAwaitingAcks() {
```
anstr, line 81: Where is access modifier(public,private,protected)?

rkatsyuryna, line 81: Package-private

PR #132, opened 2015-06-15
move from Timer to ScheduledThreadPoolExecutor

rkatsyuryna:


core/src/main/java/io/github/tcdl/Collector.java
```diff
@@ -87,7 +94,7 @@ public boolean isAwaitingResponses() {
     }
 
     public void listenForResponses(String topic, Message requestMessage) {
-        this.timer = initTimer();
+        //this.timer = initTimer();
```
anstr: Maybe this line should be removed instead of commented?

core/src/main/java/io/github/tcdl/TimerProvider.java
```diff
@@ -13,69 +16,47 @@
 
     public static final Logger LOG = LoggerFactory.getLogger(TimerProvider.class);
 
-    private TimerTask responseTimeout;
-    private TimerTask ackTimeout;
-    private Collector collector;
+    //private TimerTask responseTimeout;
```
anstr: Please remove commented lines)

rkatsyuryna: ok, agree)

YuriySavchuk:
The TimerProvider term doesn't represent functionality of the object. TimeoutManager is more better.

rkatsyuryna:
yes rename is good idea. Was thinking about this as well)

core/src/main/java/io/github/tcdl/Collector.java
```diff
@@ -40,7 +41,7 @@
     private int responsesRemaining;
 
     private Long startedAt;
-    private TimerProvider timer;
+    private TimeoutManager timer;
```
vso-tc: Looks like it's not the best name for a variable.

core/src/main/java/io/github/tcdl/TimeoutManager.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.commons.lang3.concurrent.BasicThreadFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Created by ruslan on 03.06.15.
```
vso-tc: Can we have a useful javadoc here?

core/src/test/java/io/github/tcdl/CollectorTest.java
```diff
@@ -516,8 +514,8 @@ public void testListenForResponsesEnsureResponsesRemainingIsDecreased() {
 
     private Collector initCollectorWithTimer(MsbMessageOptions messageOptions) {
 
-        Collector collector = spy(new Collector(messageOptions, msbContext, eventHandlers));
-        doReturn(timerProviderMock).when(collector).initTimer();
+        Collector collector = new Collector(messageOptions, msbContext, eventHandlers);
+        //doReturn(timeoutManagerMock).when(collector).initTimer();
```
vso-tc: why?

core/src/main/java/io/github/tcdl/Collector.java
rkatsyuryna: agree. will rename

core/src/main/java/io/github/tcdl/TimeoutManager.java
rkatsyuryna: add one

core/src/test/java/io/github/tcdl/CollectorTest.java
rkatsyuryna: removed this unnecessary method

PR #135, opened 2015-06-16
Supported extended AMQP connection parameters

YuriySavchuk:


amqp/src/main/java/io/github/tcdl/config/amqp/AmqpBrokerConfig.java
```diff
@@ -66,7 +102,8 @@ public int getConsumerThreadPoolSize() {
 
     @Override
     public String toString() {
-        return String.format("AmqpBrokerConfig [host=%s, port=%d, groupId=%s, durable=%s, consumerThreadPoolSize=%s]", host, port, groupId, durable, consumerThreadPoolSize);
+        return String.format("AmqpBrokerConfig [host=%s, port=%d, username=%s, password=%s, virtualHost=%s, groupId=%s, durable=%s, consumerThreadPoolSize=%s]",
```
vso-tc: Again, logging of passwords might not be what we want

amqp/src/main/java/io/github/tcdl/adapters/amqp/AmqpConnectionManager.java
```diff
@@ -21,15 +22,11 @@
...
-        int port = adapterConfig.getPort();
-
-        ConnectionFactory connectionFactory = new ConnectionFactory();
-        connectionFactory.setHost(host);
-        connectionFactory.setPort(port);
-
+        ConnectionFactory connectionFactory = createConnectionFactory(adapterConfig);
         try {
-            LOG.info(String.format("Opening AMQP connection to host = %s, port = %s...", host, port));
+            LOG.info(String.format("Opening AMQP connection to host = %s, port = %s, username = %s, password = %s, virtualHost = %s...",
```
vso-tc: Maybe we shouldn't log password

amqp/src/main/java/io/github/tcdl/config/amqp/AmqpBrokerConfig.java
```diff
@@ -1,20 +1,34 @@
...
 import com.typesafe.config.Config;
 
 import static io.github.tcdl.config.ConfigurationUtil.getString;
 
 public class AmqpBrokerConfig {
-
+    
     private final int port;
     private final String host;
+    private Optional<String> username;
```
vso-tc, line 14: We might have problem here because:

"The key here is the focus on use as a return type. The class is definitively not intended for use as a property of a Java Bean. Witness to this is that Optional does not implement Serializable, which is generally necessary for widespread use as a property of an object."
From http://blog.joda.org/2014/11/optional-in-java-se-8.html

amqp/src/test/java/io/github/tcdl/adapters/amqp/AmqpConnectionManagerTest.java
```diff
@@ -0,0 +1,100 @@
...
+        amqpConfig = new AmqpBrokerConfig(host, port, 
+                Optional.of(username), Optional.of(password), Optional.of(virtualHost), groupId, durable, consumerThreadPoolSize);
+        
+        try {
+            when(mockConnectionFactory.newConnection()).thenReturn(mockConnection);
+        } catch (IOException e) {
+            fail("Can't create MockConnectionFactory");
+        }
+
+        amqpConnectionManager = new AmqpConnectionManager(amqpConfig) {
```
vso-tc: Can we refactor AmqpConnectionManager to make connection factory injected and avoid modification of the class that is under test?

amqp/src/test/java/io/github/tcdl/config/amqp/AmqpBrokerConfigTest.java
```diff
@@ -13,32 +17,154 @@
...
+        String configStr = "config.amqp {"
+                + " host = \"" + host + "\"\n"
+                + " port = \"" + port + "\"\n"
+                + " groupId = \"" + groupId + "\"\n"
+                + " durable = " + durable + "\n"
+                + " consumerThreadPoolSize = " + consumerThreadPoolSize + "\n"
+                + "}";
+
+        Config amqpConfig = ConfigFactory.parseString(configStr).getConfig("config.amqp");
+        ;
```
vso-tc: extra semicolon?

core/src/main/java/io/github/tcdl/config/ConfigurationUtil.java
```diff
@@ -10,6 +12,13 @@
  */
 public class ConfigurationUtil {
 
+    public static boolean getBoolean(Config config, String key) throws ConfigurationException {
+        if (config.hasPath(key)) {
+            return config.getBoolean(key);
```
vso-tc, line 17: config.getBoolean may throw an exception. Should we handle it?

amqp/src/main/resources/amqp.conf
```diff
@@ -1,8 +1,13 @@
 config.amqp = {
 
   # AMQP Broker Adapter Defaults
+
   host = "127.0.0.1"
   port = "5672"
+  #username = ""
```
vso-tc: guest/guest is default user/password for RabbitMQ. Maybe we need to specify it explicitly here

PR #139, opened 2015-06-18
Refactoring

anstr:


core/src/main/java/io/github/tcdl/messages/Topics.java
```diff
@@ -13,32 +13,12 @@
     private final String response;
 
     @JsonCreator
-    private Topics(@JsonProperty("to") String to, @JsonProperty("response") String response) {
+    public Topics(@JsonProperty("to") String to, @JsonProperty("response") String response) {
```
vso-tc, line 16: I would define another constructor that accepts only "to" topic because "response" is optional and very often is null

anstr, line 16: The are only two places where "response" is null)And also if you add two constructors with @JsonCreator annotation you will receive java.lang.ArrayIndexOutOfBoundsException,to be honest I don't know why?!

core/src/main/java/io/github/tcdl/messages/MessageFactory.java
```diff
@@ -73,7 +73,7 @@ public MessageBuilder createBroadcastMessageBuilder(MsbMessageOptions config, St
         MetaMessageBuilder metaBuilder = createMetaBuilder(config);
         messageBuilder.setMetaBuilder(metaBuilder);
 
-        Topics topics = new Topics.TopicsBuilder().setTo(topicTo).build();
+        Topics topics = new Topics(topicTo,null);
```
wheleph: Missing space

core/src/main/java/io/github/tcdl/messages/Topics.java
wheleph, line 16: Agree, don't need to do this for only 2 places

core/src/test/java/io/github/tcdl/support/TestUtils.java
```diff
@@ -51,8 +51,7 @@ public static Message createMsbRequestMessageWithPayloadAndTopicTo(String topicT
         MsbConfigurations msbConf = createMsbConfigurations();
         Clock clock = Clock.systemDefaultZone();
 
-        Topics topic = new Topics.TopicsBuilder().setTo(topicTo)
-                .setResponse(topicTo + ":response:" + msbConf.getServiceDetails().getInstanceId()).build();
+        Topics topic = new Topics(topicTo,topicTo + ":response:" + msbConf.getServiceDetails().getInstanceId());
```
wheleph: Missing space

core/src/test/java/io/github/tcdl/support/TestUtils.java
```diff
@@ -92,8 +90,8 @@ public static Message createMsbResponseMessage() {
         MsbConfigurations msbConf = createMsbConfigurations();
         Clock clock = Clock.systemDefaultZone();
 
-        Topics topic = new Topics.TopicsBuilder().setResponse(conf.getNamespace())
-                .setTo(conf.getNamespace() + ":response:" + msbConf.getServiceDetails().getInstanceId()).build();
+        Topics topic = new Topics(conf.getNamespace(),conf.getNamespace() + ":response:" +
```
wheleph: Missing space

PR #144, opened 2015-06-19
added a configuration option to enable/disable message validation wit…

rdrozdov-tc:
…h json schema


core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -92,7 +92,7 @@ protected void handleRawMessage(String jsonMessage) {
         channelMonitorAgent.consumerMessageReceived(topic);
 
         try {
-            if (msbConfig.getSchema() != null && !Utils.isServiceTopic(topic)) {
```
rkatsyuryna, line 95: if u put first and third  if closer readability would be nicer

PR #145, opened 2015-06-19
-MsbException introduce

anstr:
-Refactoring


vso-tc:
Let's use rebase instead of merge to avoid having those "phantom" commits (Merge branch ...)

core/src/main/java/io/github/tcdl/exception/MsbException.java
```diff
@@ -0,0 +1,14 @@
+package io.github.tcdl.exception;
+
+/**
+ * Created by anstr on 6/19/2015.
```
vso-tc, line 4: Do you think this type of javadoc is useful?

PR #146, opened 2015-06-22
Removed subscriber error handler

rdrozdov-tc:


core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -107,7 +107,6 @@ protected void handleRawMessage(String jsonMessage) {
             }
         } catch (JsonConversionException | JsonSchemaValidationException e) {
             LOG.error("Got error while parsing message {}, {}", jsonMessage, e);
```
vso-tc: will it print exception stacktrace?

rdrozdov-tc: seems like jackson prints stack trace

PR #155, opened 2015-06-23
Split responses by correlation id more efficiently

rkatsyuryna:


core/src/main/java/io/github/tcdl/ChannelManager.java
```diff
@@ -52,15 +52,12 @@ public Producer findOrCreateProducer(final String topic) {
...
-    public synchronized void subscribe(final String topic, final Consumer.Subscriber subscriber) {
-        Consumer consumer = findOrCreateConsumer(topic);
-        consumer.subscribe(subscriber);
-    }
-
-    private Consumer findOrCreateConsumer(final String topic) {
+    public Consumer subscribe(String topic, Subscriber subscriber) {
         Validate.notNull(topic, "field 'topic' is null");
+        Validate.notNull(subscriber, "field 'subscriber' is null");
         Consumer consumer = consumersByTopic.computeIfAbsent(topic, key -> {
```
vso-tc, line 57: What if consumer already exist? Maybe we should explicitly throw an error

core/src/main/java/io/github/tcdl/Collector.java
```diff
@@ -96,13 +97,13 @@ public boolean isAwaitingResponses() {
     public void listenForResponses(String topic, Message requestMessage) {
         this.topic = topic;
         this.requestMessage = requestMessage;
-        channelManager.subscribe(this.topic, this);
+        collectorSubscriber.registerCollector(topic, this);
+        channelManager.subscribe(this.topic, collectorSubscriber);
```
vso-tc: looks like a hack. We overwrite the subscribe over and over again

core/src/main/java/io/github/tcdl/ChannelManager.java
rkatsyuryna, line 57: not this is ok, e.g we have few collectors that try to concurrently consume from topic. First one will win and creaae one. Others just need this ONE. Only interest is you get this ONE

core/src/main/java/io/github/tcdl/Collector.java
rkatsyuryna: no this only the name, actually subscribe is allias for "old" findOrCreateStuff - first time we create Consumer we do subscribe. All next calls will cause nothing

PR #160, opened 2015-06-24
Refactored MsbContext.

YuriySavchuk:


core/src/test/java/io/github/tcdl/RequesterTest.java
```diff
@@ -152,7 +152,8 @@ private Requester initRequesterForResponsesWithTimeout(int numberOfResponses) th
         when(channelManagerMock.findOrCreateProducer(anyString())).thenReturn(producerMock);
 
         MsbContext msbContext = TestUtils.createSimpleMsbContext();
-        msbContext.setChannelManager(channelManagerMock);
+        msbContext = TestUtils.createMsbContext(msbContext.getMsbConfig(), msbContext.getMessageFactory(),
```
vso-tc: Looks tricky. Couldn't we provide an easier way to specify channel manager instance?

core/src/test/java/io/github/tcdl/support/TestUtils.java
```diff
@@ -145,5 +150,13 @@ public static Payload createSimpleResponsePayload() {
     public static MetaMessageBuilder createSimpleMetaBuilder(MsbConfigurations msbConf, Clock clock) {
         return new MetaMessage.MetaMessageBuilder(null, clock.instant(), msbConf.getServiceDetails(), clock);
     }
+    
+    private static class TestMsbContext extends MsbContext {
```
vso-tc: With such trivial subclass couldn't we use MsbContext itself?

YuriySavchuk: Now MsbContext has only protected Constructor. And new instance can be created only by means of MsbContext.Bulder (as a usual way) or with extended class (for test purposes).

core/src/test/java/io/github/tcdl/RequesterTest.java
YuriySavchuk: MsbContext couldn't provide setters accessible for the end-user.So, setters were removed from the class. Mock implementations of the aggregated objects can be added either through Constructor directly or with using Builder pattern. The simplest approach with Constructor has been implemented.

PR #163, opened 2015-06-25
Updated configurations

YuriySavchuk:


cli/src/main/resources/application.conf
```diff
@@ -1,10 +1,33 @@
...
+
+  # Static Defaults
+  //brokerAdapterFactory = "" //Empty value equals to io.github.tcdl.adapters.mock.MockAdapterFactory
+  brokerAdapterFactory = "io.github.tcdl.adapters.amqp.AmqpAdapterFactory"
+
+  # Thread pool used for scheduling ack and response timeout tasks
+  timerThreadPoolSize: 2
+
+  # Enable/disable message validation against json schema
+  validateMessage = true
```
vso-tc: This parameter is not used in CLI tool. Let remove it because its presense is misleading

cli/src/main/resources/log4j.xml
```diff
@@ -0,0 +1,17 @@
...
+                     xmlns:log4j='http://jakarta.apache.org/log4j/'>
+
+    <appender name="consoleAppender" class="org.apache.log4j.ConsoleAppender">
+        <layout class="org.apache.log4j.PatternLayout">
+            <param name="ConversionPattern" value="%d{ABSOLUTE} %5p [%t] %c{1} - %m%n"/>
+        </layout>
+    </appender>
+
+    <root>
+        <level value="INFO"/>
```
vso-tc: I would suggest raising this to ERROR, otherwise it would be hard to find the actual messages between log messages.

PR #166, opened 2015-06-26
Javadoc for MsbContext

anstr:


core/src/main/java/io/github/tcdl/MsbContext.java
```diff
@@ -89,8 +112,11 @@ public MsbContextBuilder withDefaultChannelMonitorAgent(boolean withDefaultChann
         }
         
         /**
-         * Create MsbContext and initialize it with Config from reference.conf
-         * @throws MsbException if an error happens during initialization  
+         * Create {@link MsbContext} and initialize it with configuration from reference.conf(property file inside MSB library)
```
vso-tc, line 115: For actual microservice it's going to read application.conf as well. Doesn't it?

core/src/main/java/io/github/tcdl/MsbContext.java
```diff
@@ -46,22 +49,42 @@ public void shutdown() {
...
+     *
+     * @return factory ({@link ChannelManager}) for creating producers and consumers
+     */
     public ChannelManager getChannelManager() {
         return channelManager;
     }
 
+    /**
+     *
+     * @return object of class {@link Clock} which was crated during constructing MsbContext
```
vso-tc: "that MUST be used to obtain current time.

core/src/main/java/io/github/tcdl/MsbContext.java
```diff
@@ -15,7 +15,10 @@
 import java.time.Clock;
 
 /**
- * {@link  MsbContext} specifies the context of the MSB message processing. Creates singleton beans required for MSB.
+ * Specifies the context of the MSB message processing.
+ * Method {@link MsbContextBuilder#withShutdownHook} is used for specifying shutdown hook during JVM exit.
```
vso-tc, line 19: Why didn't you put these comments directly at the methods they describe?

anstr, line 19: At the method more detail descriptions,at the Class level briefly descriptions and link to javadoc for methods.

core/src/main/java/io/github/tcdl/MsbContext.java
anstr: Yes,this current time is time when MsbContext object was created.

core/src/main/java/io/github/tcdl/MsbContext.java
anstr, line 115: It's doesn't my line,but I will replace it with "configuration from reference.conf(property file inside MSB library) or application.conf, which will override library properties"

PR #170, opened 2015-07-02
Proposal to simplify Collector

vso-tc:


core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -33,6 +35,9 @@
     private Clock clock;
 
     private MessageHandler messageHandler;
+
+    private Map<String, MessageHandler> messageHandlersByCorrelationId;
```
vso-tc, line 39: Doesn't look right. We already tried to do it that way but this map doesn't make sense in case of ResponderServer.
Also we lose encapsulation that we had with CollectorManager

core/src/main/java/io/github/tcdl/Collector.java
```diff
@@ -135,7 +136,12 @@ protected void end() {
         cancelAckTimeoutTask();
         cancelResponseTimeoutTask();
 
-        collectorManager.unsubscribe(this);
+        Consumer consumer = channelManager.findOrCreateConsumer(topic);
```
vso-tc, line 139: It was really nice to encapsulate the unsubscription logic in CollectorManager

core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -63,6 +69,14 @@ public void subscribe(MessageHandler messageHandler) {
         this.messageHandler = messageHandler;
     }
 
+    public void addSubscriber(String correlationId, MessageHandler messageHandler) {
+        messageHandlersByCorrelationId.put(correlationId, messageHandler);
+    }
+
+    public void removeSubscriber(String correlationId) {
```
rkatsyuryna, line 76: how can we now know that last interested collector received message and we can stop consuming?

core/src/main/java/io/github/tcdl/Consumer.java
vso-tc, line 39: As far as I remember Ruslan already tried this approach

anstr, line 39: Yes this map doesn't make sense for ResponderServer,but for ResponderServer messageHandler will be used.
I don't know what do you mean by this expression "lose encapsulation",but from my point of view this proposal have the following pros:
1) We have shorter chain of method calls Collector->Consumer->Adapter instead of Collector->Consumer->CollectorManager->Adapter
2)We don't have not understandable relationship between Collector,Consumer and CollectorManager.
3)We have less numbers of lines in our code.
4)And the most important for me is that in this case we have more clear and understandable code,from my point of view.

core/src/main/java/io/github/tcdl/Consumer.java
anstr, line 76: This snippet of code is responsible for it:
    consumer.removeSubscriber(requestMessage.getCorrelationId());
        if(consumer.getMessageHandlersByCorrelationId().isEmpty()) {
            channelManager.unsubscribe(topic);
        }

This is part of end method)

core/src/main/java/io/github/tcdl/Collector.java
anstr, line 139: What benefits it will bring us?

core/src/main/java/io/github/tcdl/Collector.java
```diff
@@ -135,7 +136,12 @@ protected void end() {
...
+        consumer.removeSubscriber(requestMessage.getCorrelationId());
+        if(consumer.getMessageHandlersByCorrelationId().isEmpty()) {
+            channelManager.unsubscribe(topic);
+        }
```
rkatsyuryna, line 143: why do we need to know about getMessageHandlersByCorrelationId in Collector? this call is not obvious.

core/src/main/java/io/github/tcdl/Consumer.java
rkatsyuryna, line 76: very complex logic

anstr, line 76: This logic was moved from CollectorManager)

core/src/main/java/io/github/tcdl/Collector.java
anstr, line 143: This logic was moved from CollectorManager)And why use method getMessageHandlersByCorrelationId in Collector is bad?

core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -85,7 +99,15 @@ protected void handleRawMessage(String jsonMessage) {
             LOG.debug("Message has been successfully parsed {}", jsonMessage);
 
             if (!isMessageExpired(message)) {
-                messageHandler.handleMessage(message);
+                MessageHandler handler = messageHandlersByCorrelationId.get(message.getCorrelationId());
+                if (handler != null) {
```
vso-tc, line 103: Now we need to have these ifs to sort out the map and the handler

core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -107,7 +129,7 @@ private boolean isMessageExpired(Message message) {
         return expiryTime.isBefore(now);
     }
 
-    public MessageHandler getMessageHandler() {
-        return messageHandler;
+    public Map<String, MessageHandler> getMessageHandlersByCorrelationId() {
```
vso-tc, line 132: Breaks encapsulation. Anyone can change this collection.

PR #173, opened 2015-07-03
jbehave steps to send requests asynchronously

rdrozdov-tc:


examples/src/main/java/io/github/tcdl/examples/MSBUtil.java
```diff
@@ -14,14 +14,24 @@
 import java.util.Map;
 
 /**
- * Created by rdrozdov-tc on 6/15/15.
+ * Utility to simplify using requester and responder server
  */
-public class BaseExample {
+public class MSBUtil {
```
rkatsyuryna, line 19: can we just have MsbUtil but not MSBUtil?  Also name of class is to general, why not ComponentCreatorHelper or simple RequesterCreatorHelper (as we create only Requesters here)?

examples/src/main/java/io/github/tcdl/examples/MultipleRequesterResponder.java
```diff
@@ -11,9 +11,12 @@
 /**
  * Created by anstr on 6/9/2015.
  */
-public class MultipleRequesterResponder extends BaseExample {
+public class MultipleRequesterResponder {
+
     private static final Integer NUMBER_OF_RESPONSES = 1;
 
+    private MSBUtil util = MSBUtil.getInstance();
```
rkatsyuryna, line 18: very strange method as for Util class - getInstance . AFAIS utility classes need not instance creation

examples/src/test/resources/io/github/tcdl/examples/multiple_requests_to_date_extractor.story
```diff
@@ -0,0 +1,12 @@
+Lifecycle:
+Before:
+Given init
+And microservice DateExtractor
+After:
+Outcome: ANY
+Given shutdown
+
+Scenario: Sending multiple requests to date extractor microservice in parallel
```
rkatsyuryna, line 9: multiple is how much? can we set the number e.g 5 or 2000?

examples/src/main/java/io/github/tcdl/examples/MSBUtil.java
```diff
@@ -14,14 +14,24 @@
...
 
-    MsbContext context;
+    private static MSBUtil instance;
 
-    public BaseExample() {
-        init();
+    static {
+        instance = new MSBUtil();
```
vso-tc, line 24: Why do you need this static block?

examples/src/main/java/io/github/tcdl/examples/MSBUtil.java
```diff
@@ -14,14 +14,24 @@
...
+    }
+
+    private MsbContext context;
+
+    private MSBUtil() {
+    }
+
+    public static MSBUtil getInstance() {
+        instance.init();
```
vso-tc, line 33: weird code. When you call getInstance you overwrite msb context in the single instance you use

PR #177, opened 2015-07-03
Proposal to simplify Collector #2

vso-tc:


core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -53,6 +55,7 @@ public Consumer(ConsumerAdapter rawAdapter, String topic, MsbConfigurations msbC
         this.clock = clock;
         this.channelMonitorAgent = channelMonitorAgent;
         this.validator = validator;
+        this.collectorManager = new CollectorManager(topic, channelManager);
```
vso-tc, line 58: Now consumer has to know about Collector and CollectorManager. Why? Maybe we should put ResponderServer inside as well?
Also we have the very same problem we had before: for ResponderServer-aware consumers only messageHandler is used. Otherwise only collectorManager is used.

vso-tc, line 58: Also I think there's a bug here. collectorManager should be invoked in handleRawMessage. Shouldn't it?

anstr, line 58: > Also I think there's a bug here.

There's no bug here,see method registerCollector.

anstr, line 58: > Why? 

Consumer is used per topic,and holds collections of Collectors(inside CollectorManager) which are listening this topic.

anstr, line 58: > Maybe we should put ResponderServer inside as well?

This is sarcasm? :)

vso-tc, line 58: > There's no bug here,see method registerCollector.

You got me. `registerCollector` is too complex to me.

PR #178, opened 2015-07-05
Object factory changes

YuriySavchuk:


core/src/test/java/io/github/tcdl/impl/ResponderServerImplTest.java
```diff
@@ -63,17 +64,17 @@ public void testResponderServerProcessSuccess() throws Exception {
...
     @Test
     public void testResponderServerProcessWithError() throws Exception {
         Exception error = new Exception();
-        ResponderServerImpl.RequestHandler handler = (request, responder) -> { throw error; };
+        ResponderServer.RequestHandler handler = (request, responder) -> { throw error; };
 
-        ResponderServerImpl responderServer = (ResponderServerImpl) ResponderServerImpl
-                .create(TestUtils.getSimpleNamespace(), messageTemplate, msbContext, handler)
-                .listen();
+        ResponderServerImpl responderServer = ResponderServerImpl
```
anstr, line 74: Is it possible to use here ResponderServer responderServer = ResponderServerImpl ?

core/src/main/java/io/github/tcdl/MsbContextImpl.java
```diff
@@ -100,71 +92,8 @@ public ObjectFactory getObjectFactory() {
...
-                });
-            }
-
-            ObjectFactoryImpl objectFactory = new ObjectFactoryImpl(msbContext); 
-            msbContext.setObjectFactory(objectFactory);
-            
-            return msbContext;
-        }
-    }
 }
```
rkatsyuryna: should we move this to io.github.tcdl.impl ? Then we can easy make our RequesterImpl, ResponderImpl and ResponderServerImpl create() methods package private and allow to create this only through our MsbContext

rkatsyuryna:
would be nice to squash this two commits

core/src/test/java/io/github/tcdl/impl/ResponderServerImplTest.java
rkatsyuryna: no as we use here packege-private onResponder(). But one line above we can change:
 ResponderServerImpl.RequestHandler handler = (request, responder) -> { throw error; };
->
 ResponderServer.RequestHandler handler = (request, responder) -> { throw error; };

YuriySavchuk, line 74: We really can't change type of responderServer variable. The test verifies functionality of ResponderServerImpl class. So some of the methods (onResponder() as an example) are accessible only via ResponderServerImpl variable.

PR #179, opened 2015-07-06
added a method to public interface of responder to get original message

rdrozdov-tc:


rkatsyuryna:
what about readme.md update?

PR #180, opened 2015-07-06
Loading charset eagerly

vso-tc:


core/src/main/java/io/github/tcdl/support/JsonValidator.java
```diff
@@ -61,14 +60,14 @@ public void validate(String json, String schema) {
...
-            throw new JsonSchemaValidationException(e.getMessage());
+            throw new JsonSchemaValidationException(String.format("Error while validating message '%s' using schema '%s'", json, schema) ,e);
         }
     }
 
     public static class JsonReader {
 
-        public JsonNode read(InputStream inputStream) throws IOException{
-            return new JsonNodeReader().fromInputStream(inputStream);
+        public JsonNode read(String str) throws IOException{
```
vso-tc, line 69: Parsing JSON in more efficient and proper way. Also it prevents some nasty error messages

PR #188, opened 2015-07-07
fix consumer subscribe bug

rkatsyuryna:


core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -60,7 +59,12 @@ public Consumer(ConsumerAdapter rawAdapter, String topic, MsbConfigurations msbC
      */
     public void subscribe(MessageHandler messageHandler) {
         Validate.notNull(messageHandler, "the 'messageHandler' must not be null");
-        this.messageHandler = messageHandler;
+        if(this.messageHandler == null) {
```
vso-tc: missing whitespace

core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -60,7 +59,12 @@ public Consumer(ConsumerAdapter rawAdapter, String topic, MsbConfigurations msbC
      */
     public void subscribe(MessageHandler messageHandler) {
```
vso-tc: should it be synchronized?

core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -60,7 +59,12 @@ public Consumer(ConsumerAdapter rawAdapter, String topic, MsbConfigurations msbC
...
         Validate.notNull(messageHandler, "the 'messageHandler' must not be null");
-        this.messageHandler = messageHandler;
+        if(this.messageHandler == null) {
+            this.messageHandler = messageHandler;
+            this.rawAdapter.subscribe(this::handleRawMessage);
+        } else {
+          throw new IllegalStateException("Message handler already set for topic " + topic);
+        }
```
vso-tc: Should we set messageHandler to null in end() as well?

core/src/main/java/io/github/tcdl/Consumer.java
rkatsyuryna: find other way roung

PR #194, opened 2015-07-08
fix consumer subscribe bug

rkatsyuryna:


core/src/main/java/io/github/tcdl/collector/Collector.java
```diff
@@ -49,23 +46,30 @@
     private TimeoutManager timeoutManager;
 
     private Clock clock;
-
+    private MsbContextImpl msbContext;
```
anstr: I'm not sure that put msbContext inside Collector is good idea,as I understand MsbContext holds everything for creating Requester,ResponderServer etc..So with MsbContext  we create  Requester and it holds Collector with MsbContext as a field,very complex)
Also "private MsbContextImpl msbContext;"=>"private MsbContext msbContext;"

core/src/main/java/io/github/tcdl/collector/CollectorManagerFactory.java
```diff
@@ -0,0 +1,27 @@
+package io.github.tcdl.collector;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import io.github.tcdl.ChannelManager;
+
+public class CollectorManagerFactory {
```
anstr, line 8: Not sure that CollectorManagerFactory  is right name.

core/src/main/java/io/github/tcdl/collector/Collector.java
rkatsyuryna: removed

core/src/main/java/io/github/tcdl/collector/CollectorManagerFactory.java
rkatsyuryna, line 8: suggest better ?

core/src/main/java/io/github/tcdl/ChannelManager.java
```diff
@@ -53,10 +53,22 @@ public Producer findOrCreateProducer(final String topic) {
         return producer;
     }
 
-    public Consumer findOrCreateConsumer(final String topic) {
+    /**
+     * Start consuming messages on specified topic with handler .
+     * Calls to subscribe() and unsubscribe() have to be properly synchronized by client code not to lose messages.
```
vso-tc: I don't completely understand why do we need synchronization between subscribe and unsubscribe.

core/src/main/java/io/github/tcdl/Consumer.java
```diff
@@ -32,35 +32,31 @@
     private ChannelMonitorAgent channelMonitorAgent;
     private Clock clock;
 
+    //handler that invoked when message being received.
```
vso-tc: Let's use javadoc syntax here.

core/src/main/java/io/github/tcdl/adapters/AdapterFactory.java
```diff
@@ -12,7 +12,7 @@
     
     /**
      * Initialize AdapterFactory. The method should be called only once from AdapterFactoryLoader.
-     * @param msbConfig - MsbConfigurations object 
+     * @param msbConfig - MsbConfig object
```
vso-tc: @link would be nice here

core/src/main/java/io/github/tcdl/collector/CollectorManagerFactory.java
vso-tc, line 8: Suggesting revolution:
CollectorManager -> CollectorTopicManager
CollectorManagerFactory -> CollectorManager

core/src/test/java/io/github/tcdl/ChannelManagerConcurrentTest.java
```diff
@@ -57,6 +55,8 @@ public void testConsumerCachedMultithreadInteraction() throws Exception {
     }
 
     @Test
+    @Deprecated
```
vso-tc: It's better to leave a comment why do we ignore or deprecate some things

core/src/main/java/io/github/tcdl/collector/Collector.java
```diff
@@ -49,23 +46,25 @@
...
-    private CollectorManager collectorManager;
+    private  CollectorManager collectorManager;
+
+    public Collector(String topic, Message requestMessage, RequestOptions requestOptions, MsbContextImpl msbContext, EventHandlers eventHandlers) {
+        this.requestMessage = requestMessage;
 
-    public Collector(RequestOptions requestOptions, MsbContextImpl msbContext, EventHandlers eventHandlers) {
-        this.channelManager = msbContext.getChannelManager();
         this.clock = msbContext.getClock();
+        this.collectorManager = msbContext.getCollectorManagerFactory().findOrCreateCollectorManager(topic);
```
vso-tc, line 64: Couldn't manager be injected here? This way Collector would not create its CollectorManager

core/src/main/java/io/github/tcdl/collector/CollectorManagerFactory.java
rkatsyuryna, line 8: not sounds nicer(

core/src/main/java/io/github/tcdl/Consumer.java
rkatsyuryna: ok

core/src/main/java/io/github/tcdl/adapters/AdapterFactory.java
rkatsyuryna: ok

core/src/main/java/io/github/tcdl/collector/Collector.java
rkatsyuryna, line 64: not make sense for me, Collector constructor is already having a lot parameters.

core/src/test/java/io/github/tcdl/ChannelManagerConcurrentTest.java
rkatsyuryna: will remove this test. it's not more useful

PR #197, opened 2015-07-08
Refactoring: Renamed MsbConfigurations to MsbConfig; renamed Builders…

YuriySavchuk:
… from internal classes; updated MessageTemplate


core/src/main/java/io/github/tcdl/api/MessageTemplate.java
```diff
@@ -19,8 +23,9 @@ public Integer getTtl() {
      *
      * @param ttl time to live for the message
      */
-    public void setTtl(Integer ttl) {
+    public MessageTemplate setTtl(Integer ttl) {
```
vso-tc: maybe "withTtl"?

PR #209, opened 2015-07-14
fix MockAdapter is leaking threads

rkatsyuryna:


core/src/main/java/io/github/tcdl/msb/adapters/mock/MockAdapter.java
```diff
@@ -46,27 +55,30 @@ public void publish(String jsonMessage) {
...
-                } else {
-                    try {
-                        Thread.sleep(CONSUMING_INTERVAL);
-                    } catch (Exception e) {
-                        LOG.debug("Finish listen for subscribed topic");
+        if (executorService == null) {
+            LOG.warn("Mock Adapter not initialized for consuming");
+        } else {
+            executorService.execute(() -> {
+                {
```
vso-tc, line 66: This is dangerous because it's an infinitely running task. And ExecutorService may starve if we're not careful

vso-tc, line 66: This implementation is even more complex than before.

rkatsyuryna, line 66: no it's not infinite. we make a check for executorservice shutdown

core/src/main/java/io/github/tcdl/msb/adapters/mock/MockAdapterFactory.java
```diff
@@ -1,17 +1,25 @@
...
+import org.apache.commons.lang3.concurrent.BasicThreadFactory;
 
 /**
  * MockAdapterFactory is an implementation of {@link AdapterFactory}
  * for {@link MockAdapter}
- *
  */
 public class MockAdapterFactory implements AdapterFactory {
 
+    private static final List<ExecutorService> consumerExecutors = new LinkedList<>();
```
vso-tc: Probably we should use synchronized collection here

vso-tc: What's about removing from this list?

PR #213, opened 2015-07-15
fixed date pattern in date extractor

rdrozdov-tc:


examples/src/main/java/io/github/tcdl/msb/examples/DateExtractor.java
```diff
@@ -20,7 +20,7 @@
  */
 public class DateExtractor {
 
-    final Pattern YEAR_PATTERN = Pattern.compile("^.*(20(\\d{2})).*$");
+    final Pattern YEAR_PATTERN = Pattern.compile("^\\D*(20(\\d{2})).*$");
```
vso-tc, line 23: Could you add a unit-test for that?

PR #215, opened 2015-07-16
poc for object mapper

rdrozdov-tc:


core/src/main/java/io/github/tcdl/msb/adapters/mock/MockAdapter.java
```diff
@@ -50,9 +52,12 @@ public MockAdapter(String topic, Queue<ExecutorService> activeConsumerExecutors)
     public void publish(String jsonMessage) {
         LOG.debug("Received request {}", jsonMessage);
         try {
-            Message incomingMessage = Utils.fromJson(jsonMessage, Message.class);
-            pushRequestMessage(incomingMessage);
-        } catch (JsonConversionException e) {
+            JsonNode messageAsNode = jsonReader.read(jsonMessage);
+            if (messageAsNode.has("topics")) {
```
vso-tc: what if wit doesn't contain topics? Should we throw an exception?

rdrozdov-tc: Yes, you are right

core/src/main/java/io/github/tcdl/msb/api/MsbContextBuilder.java
```diff
@@ -108,4 +135,30 @@ public void run() {
         
         return msbContext;
     }
+
+    public ObjectMapper buildMessageMapper(ObjectMapper payloadMapper) {
```
vso-tc, line 139: Can we cover this method with unit test?

core/src/main/java/io/github/tcdl/msb/api/MsbContextBuilder.java
```diff
@@ -63,7 +78,17 @@ public MsbContextBuilder withDefaultChannelMonitorAgent(boolean withDefaultChann
         this.withDefaultChannelMonitorAgent = withDefaultChannelMonitorAgent;
         return this;
     }
-    
+
+    /**
+     * Specifies payload mapper to serialize/deserialize message payload
+     * @param payloadMapper if not provided default mapper will be used
```
rkatsyuryna: mapper seems to be not so obvious name(try to google for "java mapper"). Json ObjectMapper or  ObjectMapper will be better

acceptance/src/main/java/io/github/tcdl/msb/acceptance/MsbTestHelper.java
```diff
@@ -49,6 +51,10 @@ public MsbContext getContext() {
         return context;
     }
 
+    public ObjectMapper getObjectMapper() {
+        return ((MsbContextImpl)(getContext())).getMessageMapper();
```
vso-tc, line 55: why so many braces?

PR #216, opened 2015-07-16
Add an equals and hascode impl on Payload

sebastiaansamyn-tc:


core/src/main/java/io/github/tcdl/msb/api/message/payload/Payload.java
```diff
@@ -1,7 +1,11 @@
 package io.github.tcdl.msb.api.message.payload;
 
+import static com.google.common.base.Objects.equal;
```
vso-tc: Maybe we could use java.util.Objects instead?

PR #221, opened 2015-07-21
Fix for bug "Requester does not treat RequestOptions correctly".

anstr:


core/src/main/java/io/github/tcdl/msb/api/RequestOptions.java
```diff
@@ -43,10 +43,13 @@ public Integer getResponseTimeout() {
     }
 
     public Integer getWaitForResponses() {
-        if (waitForResponses == null || waitForResponses == -1) {
+        if (waitForResponses == null) {
             return 0;
+        } else if (waitForResponses == -1) {
+            return Integer.MAX_VALUE;
```
vso-tc: Looks like a hack. Technically we'll not be able to receive more than 5 billion responses :)

rkatsyuryna: please rename class to smth like RequesterOptions...seems like this Class has nothing to do with request itself at all(but with gathering responses in Collector)

PR #222, opened 2015-07-21
Implemeted channel monitor aggregator

vso-tc:
This is a large pull request but don't be afraid. Here're the main changes:
1. Introduced ChannelMonitorAggregator that gather stats from other microservices over the bus. It's created through object factory and is part of API.
2. Adjusted TestUtils because I need to generate new types of payloads
**3. Added method getBodyAs(TypeReference<T> typeReference) because stats payload is Map<String, AgentTopicStats>**
4. `durationMs` is not mandatory
5. Implemented small monitor service that just listens for the updates and prints them.


core/src/main/java/io/github/tcdl/msb/api/ObjectFactory.java
```diff
@@ -1,6 +1,8 @@
 package io.github.tcdl.msb.api;
 
 import io.github.tcdl.msb.api.message.Message;
+import io.github.tcdl.msb.api.monitor.AggregatorStats;
+import io.github.tcdl.msb.api.monitor.ChannelMonitorAggregator;
 
 /**
  * Provides methods for creation {@link Requester} and {@link ResponderServer}.
```
rkatsyuryna: could we mentioned as well "..creeation {@link ChannelMonitorAggregator}"?

core/src/main/java/io/github/tcdl/msb/api/message/payload/ConvertiblePayload.java
```diff
@@ -1,5 +1,7 @@
 package io.github.tcdl.msb.api.message.payload;
 
+import com.fasterxml.jackson.core.type.TypeReference;
+
 /**
  * Declares methods to convert payload parts to different types
```
rkatsyuryna, line 6: payload parts? multiple?

core/src/main/java/io/github/tcdl/msb/api/monitor/AggregatorStats.java
```diff
@@ -0,0 +1,29 @@
+package io.github.tcdl.msb.api.monitor;
+
+import io.github.tcdl.msb.config.ServiceDetails;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * This class represents internal state of {@link io.github.tcdl.msb.api.monitor.ChannelMonitorAggregator} that is passed to registered handlers.
```
rkatsyuryna: what we actually gather here? Internal state of some object sounds like a "magic". I can guess that we are interested in  Stats per Topic and per Instance

core/src/main/java/io/github/tcdl/msb/impl/ObjectFactoryImpl.java
```diff
@@ -42,4 +50,18 @@ public Requester createRequester(String namespace, RequestOptions requestOptions
...
         return ResponderServerImpl.create(namespace, messageTemplate, msbContext, requestHandler);
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public ChannelMonitorAggregator createChannelMonitorAggregator(Callback<AggregatorStats> aggregatorStatsHandler) {
+        ThreadFactory threadFactory = new BasicThreadFactory.Builder()
+                .namingPattern("monitor-aggregator-heartbeat-thread-%d")
```
rkatsyuryna, line 60: thread-%d looks not needed as we always have 1 thread

core/src/main/java/io/github/tcdl/msb/impl/ObjectFactoryImpl.java
```diff
@@ -42,4 +50,18 @@ public Requester createRequester(String namespace, RequestOptions requestOptions
     public ResponderServer createResponderServer(String namespace,  MessageTemplate messageTemplate, ResponderServer.RequestHandler requestHandler) {
         return ResponderServerImpl.create(namespace, messageTemplate, msbContext, requestHandler);
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public ChannelMonitorAggregator createChannelMonitorAggregator(Callback<AggregatorStats> aggregatorStatsHandler) {
```
rkatsyuryna, line 58: will we have  problem that after we call msbContext this monitor is still working?

core/src/main/java/io/github/tcdl/msb/monitor/TopicStats.java
```diff
@@ -1,70 +0,0 @@
-package io.github.tcdl.msb.monitor;
-
-import java.time.Instant;
-
-/**
- * Immutable class that contains statistics for a topic.
- */
-class TopicStats {
```
rkatsyuryna, line 8: immutable or effectively immutable? as i see no final here

rkatsyuryna, line 8: Also Instance statistics is mutable or not as we don't mention in JavaDoc for it

core/src/main/java/io/github/tcdl/msb/api/ObjectFactory.java
vso-tc: Thx

core/src/main/java/io/github/tcdl/msb/api/message/payload/ConvertiblePayload.java
vso-tc, line 6: Yep. Those parts are headers, params, query and body

core/src/main/java/io/github/tcdl/msb/api/monitor/AggregatorStats.java
vso-tc: Agree. I'll elaborate

core/src/main/java/io/github/tcdl/msb/impl/ObjectFactoryImpl.java
vso-tc, line 60: Defensive technique. It's better to have extra "1" in the thread name than have identical thread names if we increase the pool size in future for some reason.

core/src/main/java/io/github/tcdl/msb/impl/ObjectFactoryImpl.java
vso-tc, line 58: Thanks a very good catch. Timer will continue sending periodic heartbeats after msbContext shutdown but it will fail because the broker connection is broken.

I'll think how to fix it.

core/src/main/java/io/github/tcdl/msb/monitor/TopicStats.java
vso-tc, line 8: Effectively immutable. Thx

vso-tc, line 8: > Also Instance statistics is mutable or not as we don't mention in JavaDoc for it

?

rkatsyuryna:
Thanks a very good catch. Timer will continue sending periodic heartbeats after msbContext shutdown but it will fail because the broker connection is broken.

I'll think how to fix it.

PR #229, opened 2015-07-28
Proposal for improve subscription mechanism

anstr:


core/src/main/java/io/github/tcdl/msb/impl/ResponderServerImpl.java
```diff
@@ -47,7 +47,7 @@ static ResponderServerImpl create(String namespace,  MessageTemplate messageTemp
     public ResponderServer listen() {
         ChannelManager channelManager = msbContext.getChannelManager();
 
-        channelManager.subscribe(namespace,
+        boolean isAlreadySubscribed = channelManager.subscribe(namespace,
```
vso-tc: Why not throw exception on the lower level in ChannelManager? This solution may lead to the conflict between DefaultChannelMonitorAggregator and ResponderServer

anstr: > Why not throw exception on the lower level in ChannelManager? 

Because we will receive that Exception in method publish(RequesterImpl). It  will break functionality(more then one Collectors per topic) which we achieved with help of CollectorManager)

anstr: > This solution may lead to the conflict between DefaultChannelMonitorAggregator and ResponderServer

Agree,we can add the same flag for DefaultChannelMonitorAggregator to avoid situation when DefaultChannelMonitorAggregator and ResponderServer use the same topic.

vso-tc: Sounds very complicated. Maybe we should reconsider CollectorManager? The root cause is obviously in subscribe method and maybe we should fix it there

PR #236, opened 2015-08-06
fix unexpeced payload processing, updated JavaDoc

rkatsyuryna:


core/src/main/java/io/github/tcdl/msb/impl/ResponderServerImpl.java
```diff
@@ -60,9 +64,9 @@ public ResponderServer listen() {
     void onResponder(ResponderImpl responder) {
         Message originalMessage = responder.getOriginalMessage();
         Object rawPayload = originalMessage.getRawPayload();
-        T request = payloadMapper.convertValue(rawPayload, payloadTypeReference);
-        LOG.debug("[{}] Process message with id: [{}]", namespace, originalMessage.getId());
         try {
+            T request = Utils.convert(rawPayload, payloadTypeReference, payloadMapper);
+            LOG.debug("[{}] Process message with id: [{}]", namespace, originalMessage.getId());
```
anstr, line 69: Maybe in case Exception which is related to conservation we should send 400(Bad Request) Error Code instead of 500(Internal Server Error)?

PR #237, opened 2015-08-07
Fix responder unexpected payload. 

rkatsyuryna:
please dont merge yet untill @vso-tc can review this

### DIFF
--- a/core/src/main/java/io/github/tcdl/msb/api/monitor/AggregatorStats.java
+++ b/core/src/main/java/io/github/tcdl/msb/api/monitor/AggregatorStats.java
@@ -13,14 +13,14 @@ public class AggregatorStats {
 
     /**
      * Collects information about all topics that we know of.
-     *
+     * <p/>
      * Structure of the map: topics name -> stats
      */
     private Map<String, AggregatorTopicStats> topicInfoMap = new ConcurrentHashMap<>();
 
     /**
      * Collects information about all instances of microservices that we know of.
-     *
+     * <p/>
      * Structure of the map: service instance id -> details
      */
     private Map<String, ServiceDetails> serviceDetailsById = new ConcurrentHashMap<>();
@@ -34,10 +34,7 @@ public class AggregatorStats {
     }
 
     @Override public String toString() {
-        return "AggregatorStats{" +
-                "topicInfoMap=" + topicInfoMap +
-                ", serviceDetailsById=" + serviceDetailsById +
-                '}';
+        return String.format("AggregatorStats [topicInfoMap=%s, serviceDetailsById=%s]", topicInfoMap, serviceDetailsById);
     }
 
 }
--- a/core/src/main/java/io/github/tcdl/msb/api/monitor/AggregatorStats.java
+++ b/core/src/main/java/io/github/tcdl/msb/api/monitor/AggregatorStats.java
@@ -10,6 +10,7 @@ import java.util.concurrent.ConcurrentHashMap;
  * (that is transmitted over the bus by from their {@link io.github.tcdl.msb.monitor.agent.ChannelMonitorAgent}s).
  */
 public class AggregatorStats {
+
     /**
      * Collects information about all topics that we know of.
      *
@@ -31,4 +32,12 @@ public class AggregatorStats {
     public Map<String, ServiceDetails> getServiceDetailsById() {
         return serviceDetailsById;
     }
+
+    @Override public String toString() {
+        return "AggregatorStats{" +
+                "topicInfoMap=" + topicInfoMap +
+                ", serviceDetailsById=" + serviceDetailsById +
+                '}';
+    }
+
 }
--- a/core/src/main/java/io/github/tcdl/msb/impl/ResponderServerImpl.java
+++ b/core/src/main/java/io/github/tcdl/msb/impl/ResponderServerImpl.java
@@ -8,6 +8,7 @@ import io.github.tcdl.msb.api.Responder;
 import io.github.tcdl.msb.api.ResponderServer;
 import io.github.tcdl.msb.api.message.Message;
 import io.github.tcdl.msb.api.message.payload.Payload;
+import io.github.tcdl.msb.support.Utils;
 import org.apache.commons.lang3.Validate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -41,7 +42,10 @@ public class ResponderServerImpl<T extends Payload> implements ResponderServer<T
     }
 
     /**
-     * {@inheritDoc}
+     * Start listening for message on specified topic.
+     *
+     * In case of exception was thrown during message conversion or handling business logic
+     * response message with statusCode=500 and statusMessage=${exception.getMessage()} will be created and send automatically.
      */
     @Override
     public ResponderServer listen() {
@@ -60,9 +64,9 @@ public class ResponderServerImpl<T extends Payload> implements ResponderServer<T
     void onResponder(ResponderImpl responder) {
         Message originalMessage = responder.getOriginalMessage();
         Object rawPayload = originalMessage.getRawPayload();
-        T request = payloadMapper.convertValue(rawPayload, payloadTypeReference);
-        LOG.debug("[{}] Process message with id: [{}]", namespace, originalMessage.getId());
         try {
+            T request = Utils.convert(rawPayload, payloadTypeReference, payloadMapper);
+            LOG.debug("[{}] Process message with id: [{}]", namespace, originalMessage.getId());
             requestHandler.process(request, responder);
         } catch (Exception exception) {
             errorHandler(responder, exception);
--- a/core/src/main/java/io/github/tcdl/msb/impl/ResponderServerImpl.java
+++ b/core/src/main/java/io/github/tcdl/msb/impl/ResponderServerImpl.java
@@ -45,7 +45,7 @@ public class ResponderServerImpl<T extends Payload> implements ResponderServer<T
      * Start listening for message on specified topic.
      *
      * In case of exception was thrown during message conversion or handling business logic
-     * response message with statusCode=500 and statusMessage=${exception.getMessage()} will be created and send automatically.
+     * response message with statusCode=500 and statusMessage=${exception.getMessage()} will be created and sent automatically.
      */
     @Override
     public ResponderServer listen() {
--- a/core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
+++ b/core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
@@ -71,13 +71,18 @@ public class DefaultChannelMonitorAggregator implements ChannelMonitorAggregator
     void onHeartbeatResponses(List<Message> heartbeatResponses) {
         LOG.debug(String.format("Handling heartbeat responses %s...", heartbeatResponses));
         AggregatorStats aggregatorStats = new AggregatorStats();
+        boolean successfullyAggregatedAtLeastOne = false;
         for (Message msg : heartbeatResponses) {
-            aggregateInfo(aggregatorStats, msg);
+            if (aggregateInfo(aggregatorStats, msg)) {
+                successfullyAggregatedAtLeastOne = true;
+            }
         }
-        LOG.debug(String.format("Calling registered handler for aggregated statistics %s...", masterAggregatorStats));
-        handler.call(aggregatorStats);
-        masterAggregatorStats = aggregatorStats;
-        LOG.debug(String.format("Heartbeat responses processed"));
+        if (successfullyAggregatedAtLeastOne) {
+            LOG.debug(String.format("Calling registered handler for heartbeat statistics %s...", masterAggregatorStats));
+            handler.call(aggregatorStats);
+            masterAggregatorStats = aggregatorStats;
+            LOG.debug(String.format("Heartbeat responses processed"));
+        }
     }
 
     void onAnnounce(Message announcementMessage) {
@@ -85,12 +90,10 @@ public class DefaultChannelMonitorAggregator implements ChannelMonitorAggregator
 
         boolean successfullyAggregated = aggregateInfo(masterAggregatorStats, announcementMessage);
 
-        if(successfullyAggregated) {
-            LOG.debug(String.format("Calling registered handler for for aggregated statistics %s...", masterAggregatorStats));
+        if (successfullyAggregated) {
+            LOG.debug(String.format("Calling registered handler for announcement statistics %s...", masterAggregatorStats));
             handler.call(masterAggregatorStats);
             LOG.debug(String.format("Announcement message processed"));
-        } else {
-            LOG.error("Message [{}] will be skipped in statistics computation.", announcementMessage);
         }
     }
 
@@ -98,7 +101,7 @@ public class DefaultChannelMonitorAggregator implements ChannelMonitorAggregator
 
         JsonNode rawPayload = message.getRawPayload();
 
-        if(!Utils.isPayloadPresent(rawPayload)) {
+        if (!Utils.isPayloadPresent(rawPayload)) {
             LOG.error("Unable to convert message. Message payload is empty.");
             return false;
         }
@@ -119,7 +122,7 @@ public class DefaultChannelMonitorAggregator implements ChannelMonitorAggregator
             return false;
         }
 
-       return true;
+        return true;
     }
 
     void aggregateTopicStats(AggregatorStats aggregatorStats, Map<String, AgentTopicStats> agentTopicStatsMap, String instanceId) {
--- a/core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
+++ b/core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
@@ -1,10 +1,19 @@
 package io.github.tcdl.msb.monitor.aggregator;
 
+import static io.github.tcdl.msb.support.Utils.TOPIC_ANNOUNCE;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.github.tcdl.msb.ChannelManager;
 import io.github.tcdl.msb.api.Callback;
 import io.github.tcdl.msb.api.ObjectFactory;
+import io.github.tcdl.msb.api.exception.JsonConversionException;
 import io.github.tcdl.msb.api.message.Message;
 import io.github.tcdl.msb.api.message.MetaMessage;
 import io.github.tcdl.msb.api.message.payload.Payload;
@@ -14,16 +23,9 @@ import io.github.tcdl.msb.api.monitor.ChannelMonitorAggregator;
 import io.github.tcdl.msb.config.ServiceDetails;
 import io.github.tcdl.msb.impl.MsbContextImpl;
 import io.github.tcdl.msb.monitor.agent.AgentTopicStats;
+import io.github.tcdl.msb.support.Utils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.util.List;
-import java.util.Map;
-import java.util.Map.Entry;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
-
-import static io.github.tcdl.msb.support.Utils.TOPIC_ANNOUNCE;
 
 public class DefaultChannelMonitorAggregator implements ChannelMonitorAggregator {
     private static final Logger LOG = LoggerFactory.getLogger(DefaultChannelMonitorAggregator.class);
@@ -36,7 +38,8 @@ public class DefaultChannelMonitorAggregator implements ChannelMonitorAggregator
 
     AggregatorStats masterAggregatorStats = new AggregatorStats();
 
-    public DefaultChannelMonitorAggregator(MsbContextImpl msbContext, ScheduledExecutorService scheduledExecutorService, Callback<AggregatorStats> aggregatorStatsHandler) {
+    public DefaultChannelMonitorAggregator(MsbContextImpl msbContext, ScheduledExecutorService scheduledExecutorService,
+            Callback<AggregatorStats> aggregatorStatsHandler) {
         this.channelManager = msbContext.getChannelManager();
         this.objectFactory = msbContext.getObjectFactory();
         this.messageMapper = msbContext.getPayloadMapper();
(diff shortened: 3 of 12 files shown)


rkatsyuryna:
not sure about use of DefaultChannelMonitorAggregator.masterAggregatorStats. 
Expect we put (aggregate) statistics during whole monitor running time. 
But in DefaultChannelMonitorAggregator.onHeartbeatResponses we create stats from scratch and reasign to masterAggregatorStats.

core/src/main/java/io/github/tcdl/msb/api/monitor/AggregatorStats.java
```diff
@@ -31,4 +32,12 @@
     public Map<String, ServiceDetails> getServiceDetailsById() {
         return serviceDetailsById;
     }
+
+    @Override public String toString() {
```
vso-tc, line 36: Missing newline here.

core/src/main/java/io/github/tcdl/msb/impl/ResponderServerImpl.java
```diff
@@ -41,7 +42,10 @@ private ResponderServerImpl(String namespace, MessageTemplate messageTemplate, M
     }
 
     /**
-     * {@inheritDoc}
+     * Start listening for message on specified topic.
+     *
+     * In case of exception was thrown during message conversion or handling business logic
+     * response message with statusCode=500 and statusMessage=${exception.getMessage()} will be created and send automatically.
```
vso-tc: send -> sent

core/src/main/java/io/github/tcdl/msb/impl/ResponderServerImpl.java
```diff
@@ -60,9 +64,9 @@ public ResponderServer listen() {
     void onResponder(ResponderImpl responder) {
         Message originalMessage = responder.getOriginalMessage();
         Object rawPayload = originalMessage.getRawPayload();
-        T request = payloadMapper.convertValue(rawPayload, payloadTypeReference);
-        LOG.debug("[{}] Process message with id: [{}]", namespace, originalMessage.getId());
         try {
+            T request = Utils.convert(rawPayload, payloadTypeReference, payloadMapper);
```
vso-tc, line 68: Good catch! With 2-step message parsing we're able to do such trick.

core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
```diff
@@ -71,31 +74,52 @@ void onHeartbeatResponses(List<Message> heartbeatResponses) {
...
     void onAnnounce(Message announcementMessage) {
         LOG.debug(String.format("Handling announcement message %s...", announcementMessage));
-        aggregateInfo(masterAggregatorStats, announcementMessage);
-        LOG.debug(String.format("Calling registered handler for announcement %s...", announcementMessage));
-        handler.call(masterAggregatorStats);
-        LOG.debug(String.format("Announcement message processed %s", announcementMessage));
+
+        boolean successfullyAggregated = aggregateInfo(masterAggregatorStats, announcementMessage);
+
+        if(successfullyAggregated) {
```
vso-tc: missing space

core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
```diff
@@ -71,31 +74,52 @@ void onHeartbeatResponses(List<Message> heartbeatResponses) {
...
-        String instanceId = serviceDetails.getInstanceId();
-        aggregatorStats.getServiceDetailsById().put(instanceId, serviceDetails);
+    boolean aggregateInfo(AggregatorStats aggregatorStats, Message message) {
+
+        JsonNode rawPayload = message.getRawPayload();
 
-        Object rawPayload = message.getRawPayload();
-        Payload<?, ?, ?, Map<String, AgentTopicStats>> payload = messageMapper.convertValue(rawPayload, new TypeReference<Payload<Object, Object, Object, Map<String, AgentTopicStats>>>() {});
-        Map<String, AgentTopicStats> agentTopicStatsMap = payload.getBody();
+        if(!Utils.isPayloadPresent(rawPayload)) {
```
vso-tc: missing space

core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
```diff
@@ -71,31 +74,52 @@ void onHeartbeatResponses(List<Message> heartbeatResponses) {
         for (Message msg : heartbeatResponses) {
             aggregateInfo(aggregatorStats, msg);
         }
-        LOG.debug(String.format("Calling registered handler for heartbeat %s...", heartbeatResponses));
+        LOG.debug(String.format("Calling registered handler for aggregated statistics %s...", masterAggregatorStats));
```
vso-tc: I don't see the reason to have identical log messages for announcements and heartbeats. Different messages might ease problem investigation and don't force you to copy/paste them.

If you really want to make the identical I would suggest to extract LOG.debug and handler.call into a separate method and call it twice.

core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
```diff
@@ -71,31 +74,52 @@ void onHeartbeatResponses(List<Message> heartbeatResponses) {
         for (Message msg : heartbeatResponses) {
             aggregateInfo(aggregatorStats, msg);
```
vso-tc: You ignore the return value of aggregateInfo here. Looks fishy.
If you're unable to process one heartbeat response then you aggregatorStats might not be accurate but you still call the handler. 

It might be OK but I would argue that at least logging should be similar to "onAnnounce". There you log error message when payload is not valid so why don't you doing it here? In fact I wouldn't log anything in both cases because "aggregateInfo" already have a very detailed logging that explains clearly why the message is considered to be invalid

vso-tc: And yeah, I can explain you why onAnnonce adds the response while onHeartbeatResponses overwrites it. The thing is that this aggregator can work in 2 modes (this logic is taken from Node msb): 
- listen only that only accepts announcements but doesn't send periodic heartbeats
- above + sending periodic heartbeats. In this case we want to overwrite stats with most recent information to detect that some microservices went down.

core/src/main/java/io/github/tcdl/msb/support/Utils.java
```diff
@@ -56,8 +58,8 @@ public static String toJson(Object object, ObjectMapper objectMapper) {
         try {
             return objectMapper.writeValueAsString(object);
         } catch (JsonProcessingException e) {
-            LOG.error("Failed parse to JSON from: " + object, e);
-            throw new JsonConversionException(e.getMessage());
+            LOG.error("Failed to parse to JSON object: [{}] ", object);
```
vso-tc, line 61: I know that you didn't change this code but "log & throw" is considered to be evil: http://stackoverflow.com/questions/6639963/why-is-log-and-throw-condsidered-an-anti-pattern

In our case we'll end up logging the same parsing issue from multiple catch blocks in different methods. Could we get rid of it?

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
```diff
@@ -37,10 +49,143 @@ public void testAnnouncement() throws InterruptedException {
...
+
+        //simulate unexpectedMessage was consumed from in TOPIC_ANNOUNCE namespace
+        MockAdapter.pushRequestMessage(Utils.TOPIC_ANNOUNCE,
+                Utils.toJson(TestUtils.createMsbRequestMessageWithSimplePayload(Utils.TOPIC_ANNOUNCE), msbContext.getPayloadMapper()));
+
+        ChannelMonitorAggregator channelMonitorAggregator = msbContext.getObjectFactory().createChannelMonitorAggregator(handler);
+        channelMonitorAggregator.start(false, ChannelMonitorAggregator.DEFAULT_HEARTBEAT_INTERVAL_MS, ChannelMonitorAggregator.DEFAULT_HEARTBEAT_TIMEOUT_MS);
+        channelMonitorAgent.producerTopicCreated(TOPIC_NAME);
+
+        assertTrue("Announcement was not received", announcementReceived.await(RequesterResponderIT.MESSAGE_TRANSMISSION_TIME * 2, TimeUnit.MILLISECONDS));
```
vso-tc: That's a dangerous assert might be very fragile.

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
```diff
@@ -37,10 +49,143 @@ public void testAnnouncement() throws InterruptedException {
...
+    }
+
+    @Test
+    public void testHeartbeatMessage() throws InterruptedException {
+        String TOPIC_NAME = "topic2";
+
+        Map<String, AgentTopicStats> topicInfoMap = new HashMap<>();
+        Instant lastProduced = Instant.now().minusMillis(20000);
```
vso-tc: Looks dangerous. I would suggest using fixed clock here to avoid problems with slow Travis. In fact you could use constants here, why bother with subtractions?

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
```diff
@@ -37,10 +49,143 @@ public void testAnnouncement() throws InterruptedException {
...
+                Thread.sleep(100);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+            }
+        }
+        Message requestMessage = Utils.fromJson(originalRequest, Message.class, msbContext.getPayloadMapper());
+        return requestMessage;
+    }
+
+    private Message createMsbRequestMessage(String topicTo, String correlationId, String payloadString) {
```
vso-tc, line 186: This helper should be in TestUtils. BTW, can't we adjust and reuse some method from there?

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
```diff
@@ -37,10 +49,143 @@ public void testAnnouncement() throws InterruptedException {
...
+                    .withTopics(topic)
+                    .withMetaBuilder(metaBuilder)
+                    .withPayload(payload)
+                    .build();
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to prepare request message", e);
+        }
+    }
+
+    private Message createMsbRequestMessage(String topicTo, String correlationId, Payload payload) {
```
vso-tc, line 208: This helper should be in TestUtils. BTW, can't we adjust and reuse some method from there?

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
```diff
@@ -37,10 +49,143 @@ public void testAnnouncement() throws InterruptedException {
...
+        MockAdapter.pushRequestMessage(requestMessage.getTopics().getResponse(), Utils.toJson(responseMessage, msbContext.getPayloadMapper()));
+
+        assertTrue("HeartBeat Response was not received",
+                heartBeatResponseReceived.await(RequesterResponderIT.MESSAGE_TRANSMISSION_TIME, TimeUnit.MILLISECONDS));
+    }
+
+    private Message awaitHeartBeatRequestSent(long timeout) {
+        long startTime = System.currentTimeMillis();
+        String originalRequest = null;
+        while (originalRequest == null && (System.currentTimeMillis() - startTime) < timeout) {
```
vso-tc: We agreed not to use System.currentTimeMillis but use java.time instead.

vso-tc: Also why do you have this loop here? It seems like you re-implement the consumer.

Could you just hook up the real consumer to the hearbet topic?

core/src/main/java/io/github/tcdl/msb/api/monitor/AggregatorStats.java
rkatsyuryna, line 36: ah, Idea generate this for me((. Will fix

core/src/main/java/io/github/tcdl/msb/impl/ResponderServerImpl.java
rkatsyuryna: fixed

core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
rkatsyuryna: 1st comment - >
"If you're unable to process one heartbeat response then you aggregatorStats might not be accurate but you still call the handler." 
- YES this is true in case heartbeat responses and that's why i ignore return value.
  I will FIX  -> check for at least ONE response was OK and call handler. This sounds even more clean approach. Update IT test for that.
  -Logging. Agree. will remove extra form onAnnounce().  Not sure if need to log case when  non of heartbeatsResponses where aggregated. Any ideas?

2-nd comment ->
cool, it would be nice to describe this "additional feature" about heartbeats in javadoc. Will add info in ChannelMonitorAggregator class Javadoc. Please take a look

core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
rkatsyuryna: agree. Fixed

core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
rkatsyuryna: hard to say where. but tried my best to fix

core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
rkatsyuryna: hard to say where. but tried my best to fix

core/src/main/java/io/github/tcdl/msb/support/Utils.java
rkatsyuryna, line 61: I would say i improved a lot. We log only in Utils ->
Failed to parse from JSON: [{}] to object of type:
Failed to parse to JSON object: [{}]
Failed to convert object [{}] to type:
So Utils logs context it knows about -> objects, raw json and most important - type we convert into.

Actual callers to this are Producer/Consumer/DefaultChannelMonitorAggregator (but this is runtime so we not insist on this as result can't expect they will log or smth) just catch JsonConversionException and log exception with context they know. 

Example in DefaultChannelMonitorAggregator log:

ERROR [timer-provider-thread-1] Utils Failed to convert object [{"body":" unexpected statistics format received"}] to type: [io.github.tcdl.msb.api.message.payload.Payload<java.lang.Object, java.lang.Object, java.lang.Object, java.util.Map<java.lang.String, io.github.tcdl.msb.monitor.agent.AgentTopicStats>>]

12:35:31,362 ERROR [timer-provider-thread-1] DefaultChannelMonitorAggregator Unable to convert message.
io.github.tcdl.msb.api.exception.JsonConversionException: Can not instantiate value of type java.util.LinkedHashMap from String value (' unexpected statistics format received'); no single-String constructor/factory method
 at [Source: N/A; line: -1, column: -1]
    at io.github.tcdl.msb.support.Utils.convert(Utils.java:107)

Example in ResponderServerImpl:
12:41:07,292 ERROR [main] Utils Failed to convert object [{"body":"some body"}] to type: [io.github.tcdl.msb.api.message.payload.Payload<?, ?, ?, java.lang.Integer>]
12:41:07,292 ERROR [main] ResponderServerImpl [test:responder-server] Error while processing message with id: [090c2bfa-7ca5-4457-bde2-c23fd7e639ef]. Cause: [Can not construct instance of int from String value 'some body': not a valid Integer value
 at [Source: N/A; line: -1, column: -1]]

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
rkatsyuryna: why. I don't see a chance for it now. 
Added 
 @After
    public void tearDown() {
        channelMonitorAggregator.stop();
    }

So no chance for problems :)

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
rkatsyuryna: this is just information in POJO. It can be any value as it's purpose is simple to send > parse

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
rkatsyuryna, line 186: we need to clear methods in TestUtils before adding new there (there are too much there and if ad more it's gonna be total mess ). Think of it as a separate task, Not in scope of this

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
rkatsyuryna, line 208: we need to clear methods in TestUtils before adding new there (there are too much there and if ad more it's gonna be total mess ). Think of it as a separate task, Not in scope of this

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
rkatsyuryna: 1-st -> fixed
2-nd ->
i don't want to use consumer from here. Think about this just as overload for  MockAdapter.pollJsonMessageForTopic.  Move that to MockAdapter directly?. 

I need this loop to simulate response for heartbeat request with same CorrelationId. See now other way to achieve this.

core/src/main/java/io/github/tcdl/msb/support/Utils.java
vso-tc, line 61: I see your point and it's a valid one. Though it's somewhat hard to relate those messages. Maybe we could twist in message id in some way to be available for responder server? Like: http://skipy.ru/useful/logging.html#log4j_ndc

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
vso-tc: Yep, that's why there's no need to make it more complicated that it should be. If we're talking about POJO let's use constants here. That would be more safe and expressive

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
vso-tc, line 186: Then we should at least put TODO comment here not to forget about this test once we get to that other task.

core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
```diff
@@ -68,34 +71,58 @@ public void stop() {
     void onHeartbeatResponses(List<Message> heartbeatResponses) {
         LOG.debug(String.format("Handling heartbeat responses %s...", heartbeatResponses));
         AggregatorStats aggregatorStats = new AggregatorStats();
+        boolean successfullyAggregatedAtLeastOne = false;
         for (Message msg : heartbeatResponses) {
-            aggregateInfo(aggregatorStats, msg);
+            if (aggregateInfo(aggregatorStats, msg) == true) {
```
vso-tc: explicit "== true" looks weird.

core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
```diff
@@ -68,34 +71,58 @@ public void stop() {
...
+                successfullyAggregatedAtLeastOne = true;
+            }
```
vso-tc, line 78: consier getting rid of the if statement:

``` java
boolean successfullyAggregated = aggregateInfo(aggregatorStats, msg);
successfullyAggregatedAtLeastOne |= successfullyAggregated;
```

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
```diff
@@ -37,10 +58,158 @@ public void testAnnouncement() throws InterruptedException {
...
+
+        //simulate unexpectedMessage was consumed from in TOPIC_ANNOUNCE namespace
+        MockAdapter.pushRequestMessage(Utils.TOPIC_ANNOUNCE,
+                Utils.toJson(TestUtils.createMsbRequestMessageWithSimplePayload(Utils.TOPIC_ANNOUNCE), msbContext.getPayloadMapper()));
+
+        channelMonitorAggregator = msbContext.getObjectFactory().createChannelMonitorAggregator(handler);
+        channelMonitorAggregator.start(false, ChannelMonitorAggregator.DEFAULT_HEARTBEAT_INTERVAL_MS, ChannelMonitorAggregator.DEFAULT_HEARTBEAT_TIMEOUT_MS);
+        channelMonitorAgent.producerTopicCreated(TOPIC_NAME);
+
+        assertTrue("Announcement was not received", announcementReceived.await(RequesterResponderIT.MESSAGE_TRANSMISSION_TIME * 2, TimeUnit.MILLISECONDS));
```
vso-tc: Is it a valid test? You're sending 2 messages and waiting for only 1 of them to be consumed. But await is unblocked until ANY of the responses arrive. 

So let's consider a case where our code is broken so that the handler is invoked even for an invalid message. In this scenario the verificaion will still pass without waiting for the second message. Am I missing something?

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
```diff
@@ -37,10 +58,158 @@ public void testAnnouncement() throws InterruptedException {
...
+        MockAdapter.pushRequestMessage(requestMessage.getTopics().getResponse(), Utils.toJson(brokenResponseMessage, msbContext.getPayloadMapper()));
+
+        assertTrue("Heartbeat response was not received",
+                heartBeatResponseReceived.await(RequesterResponderIT.MESSAGE_TRANSMISSION_TIME, TimeUnit.MILLISECONDS));
+    }
+
+    private Message awaitHeartBeatRequestSent(long timeoutMs) {
+        LocalDateTime startTime = LocalDateTime.now();
+        String originalRequest = null;
+        while (originalRequest == null && (LocalDateTime.now().isBefore(startTime.plus(timeoutMs, ChronoUnit.MILLIS)))) {
```
vso-tc: This line is hard to read. Let's make it shorter by using local vars

vso-tc: Again, why can't you use consumer here?

core/src/main/java/io/github/tcdl/msb/support/Utils.java
rkatsyuryna, line 61: I don't see the reason for that. 
In case i want to log messageId we need -> 
1.pass messageId to ndc + pop messageId in Utils. 
2.still log this in Producer/Consumer/DefaultChannelMonitorAggregator as we need context where this happened.

Is that what we want? or do i miss smth?

p.s. If vice versa pass context from Utils to Producer/Consumer/DefaultChannelMonitorAggregator we need to be sure after we push in Utils other component will have to pop. If we add new that use Utils that can brake and get even worth.

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
rkatsyuryna: Constant...no problem. Fix this

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
rkatsyuryna, line 186: added

core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
rkatsyuryna: well some people say opposite)) anyway will get rid of.

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
rkatsyuryna: Test is 100% valid.
no way, we are waiting not for Response but End event..but yes in some way if you broke  our end event logic this will pass. But this is all i can do here

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
rkatsyuryna: 1st - don't understand.
2 extension to MockAdapter. We dont use any cosnumer-requester there..just operate on Maps directly.this is just a version of MockAdapter.pollJsonMessageForTopic with await.

core/src/main/java/io/github/tcdl/msb/monitor/aggregator/DefaultChannelMonitorAggregator.java
rkatsyuryna, line 78: this comment i can't understand. sorry

core/src/main/java/io/github/tcdl/msb/impl/ResponderServerImpl.java
```diff
@@ -60,9 +64,9 @@ public ResponderServer listen() {
...
+            LOG.debug("[{}] Process message with id: [{}]", namespace, originalMessage.getId());
```
anstr, line 69: Is it correct to send 500 Internal Error?Are we agreed on it with our architects?From my side in this case we should use 400 Bad Request)

rkatsyuryna, line 69: Totally agree.  But don't wont to block now because of that. Will be discussed this later

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
rkatsyuryna: fixed after deep analysis and clarification by @vso-tc

core/src/test/java/io/github/tcdl/msb/api/ChannelMonitorIT.java
rkatsyuryna: revrite after deep analysis and clarification by @vso-tc

PR #240, opened 2015-08-12
Made javac output all warnings. Fixed the warnings

vso-tc:


pom.xml
```diff
@@ -158,7 +158,7 @@
                     <source>1.8</source>
                     <target>1.8</target>
                     <compilerArgs>
-                        <arg>-Xlint:unchecked</arg>
+                        <arg>-Xlint:all</arg>
```
vso-tc, line 161: Made javac output all the warnings

PR #243, opened 2015-08-14
Fix for FacetsAggregator

anstr:


examples/src/main/java/io/github/tcdl/msb/examples/FacetsAggregator.java
```diff
@@ -11,17 +11,14 @@
...
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 /**
  * Microservice which is listening for incoming messages, creates requests to another microservices(
  * data-extractor, airport-extractor, resort-extractor), concatenates responses and returns result response
  */
 public class FacetsAggregator {
+    static List<Payload> responses = new LinkedList<>();
```
vso-tc: Is it thread-safe? Seems like the instance is shared between all requesters and is never cleared

PR #246, opened 2015-08-18
Convert Message.bodyBuffer from String to byte array

rkatsyuryna:


core/src/test/java/io/github/tcdl/msb/api/RequesterIT.java
```diff
@@ -35,6 +36,17 @@ public void testRequestMessage() throws Exception {
         TestUtils.assertRequestMessagePayload(adapterJsonMessage, requestPayload, NAMESPACE);
     }
 
+    @Test()
```
vso-tc: extra "()"

core/src/test/java/io/github/tcdl/msb/api/RequesterIT.java
```diff
@@ -35,6 +36,17 @@ public void testRequestMessage() throws Exception {
...
+    public void testRequestMessageWithBodyBufferBase64Encoded() throws Exception {
+        Payload requestPayload = TestUtils.createSimpleRequestPayloadWithBodyBuffer(new byte[24]);
```
vso-tc: I would create simple test that uses 1 or 2 bytes to make it more expressive

core/src/test/java/io/github/tcdl/msb/api/RequesterIT.java
rkatsyuryna: sure

core/src/test/java/io/github/tcdl/msb/api/RequesterIT.java
rkatsyuryna: tried to make same as in node.js version. but will update

PR #247, opened 2015-08-19
added tags to few examples

rdrozdov-tc:


examples/src/main/java/io/github/tcdl/msb/examples/PingService.java
```diff
@@ -32,6 +35,6 @@ public static void main(String[] args) {
                 .withBody("PING")
                 .build();
 
-        requester.publish(pingPayload); // Send the message
+        requester.publish(pingPayload, "ping-dynamic-tag"); // Send the message
```
vso-tc: It's not that dynamic :)

examples/src/main/java/io/github/tcdl/msb/examples/FacetsAggregator.java
```diff
@@ -30,7 +30,7 @@ public static void main(String[] args) throws ScriptException, FileNotFoundExcep
                 .enableShutdownHook(true)
                 .build();
 
-        MessageTemplate options = new MessageTemplate();
+        MessageTemplate options = new MessageTemplate().withTags("facets-aggregator");
         final String namespace = "search:aggregator:facets:v1";
```
vso-tc, line 36: I would say that we need to have dynamic tags in this example as well

PR #253, opened 2015-08-27
Implemented certificate-less SSL support in master

vso-tc:


amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpAdapterFactory.java
```diff
@@ -96,6 +94,19 @@ protected ConnectionFactory createConnectionFactory(AmqpBrokerConfig adapterConf
         if(virtualHost.isPresent()) {
             connectionFactory.setVirtualHost(virtualHost.get());
         }
+
+        try {
+            if (adapterConfig.useSSL()) {
+                LOG.info("Configuring adapter to use SSL...");
```
rkatsyuryna, line 100: maybe debug level? two almost identical messages

vso-tc, line 100: I would still use info. These log statements trace the process of application initialization. There's no harm in having 2 messages performance-wise bit it may simplify investigation of a problem

amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpAdapterFactory.java
```diff
@@ -96,6 +94,19 @@ protected ConnectionFactory createConnectionFactory(AmqpBrokerConfig adapterConf
...
+                connectionFactory.useSslProtocol();
```
rkatsyuryna, line 101: this use DEFAULT_SSL_PROTOCOL. 
From this http://grepcode.com/file/repo1.maven.org/maven2/com.rabbitmq/amqp-client/3.5.3/com/rabbitmq/client/ConnectionFactory.java#ConnectionFactory.0DEFAULT_SSL_PROTOCOL

i see that it's hardcoded to  `private static final String DEFAULT_SSL_PROTOCOL = "TLSv1"`

Are we ok with that? Newer protocols are TLS 1.1 and TLS 1.2.

vso-tc, line 101: I think we are. 

This change was intended as a quick fix the the blocker Dragons have. Their server uses TLSv1. Also this is the default value in amqp client which suggests that it should probably be good enough for most cases.

I agree that for full SSL support we should allow developers to specify the encryption they would like to use. But that should be a part of separate story for implementation of full SSL support in msb-java (including certificate validation).

rkatsyuryna, line 101: ok

PR #260, opened 2015-09-04
Moved InstanceId from mandatory to optional parameter.

YuriySavchuk:


core/src/test/java/io/github/tcdl/msb/config/ServiceDetailsTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.junit.Test;
+
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+public class ServiceDetailsTest {
+    final String name = "test_msb";
+    final String version = "1.0.1";
+    final String instanceId = "msbd06a-ed59-4a39-9f95-811c5fb6ab87";
```
anstr, line 16: How did you get this UUID?

core/src/test/java/io/github/tcdl/msb/config/ServiceDetailsTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Test
+    public void testServiceDetailsAll() {
+        String serviceDetailsConfigStr = "serviceDetails = {name = \"" + name + "\", version = \"" + version + "\", instanceId = \"" + instanceId + "\"}";
```
vso-tc: I would suggest using String.format instead of string concatenation for readability reasons

core/src/test/java/io/github/tcdl/msb/config/ServiceDetailsTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+        assertEquals("expect \"" + name + "\" as a name value", name, serviceDetails.getName());
+        assertEquals("expect \"" + version + "\" as a vesrion value", version, serviceDetails.getVersion());
+        assertNotNull("expect InstanceId is not null", serviceDetails.getInstanceId());
+
+        String instanceId1 = serviceDetails.getInstanceId();
+        String instanceId2 = serviceDetails.getInstanceId();
+        assertEquals("expect stabel InstanceId value", instanceId1, instanceId2);
+    }
+
```
vso-tc, line 62: I would add another test that verifies that when we parse configuration without instanceId twice, the generated ids are different.

PR #263, opened 2015-09-07
Added description for empty Instance Id (WEB-12810).

YuriySavchuk:


doc/MSB.md
```diff
@@ -282,11 +282,13 @@ All configuration files use _key-value pair_ structure.
 ### Description of MSB configuration fields
 Service details section describes microservice parameters.
 
-- `name ` – microservice name
-- `version` – microservice version
-- `instanceId` – unique microservice instance id
+- `name ` – microservice name. All run instances of the same microservice must have the same name. 
+- `version` – microservice version. At the moment Msb-java doesn't handle the value.
+- `instanceId` – unique microservice instance id. All run instances of the same microservice should have different instance ids.
```
vso-tc: Run -> Running?

Maybe let's explicitly specify that after microservice restart the value should be different. Also let's mention how this value is used (to construct reply topic name)

doc/MSB.md
```diff
@@ -282,11 +282,13 @@ All configuration files use _key-value pair_ structure.
 ### Description of MSB configuration fields
 Service details section describes microservice parameters.
 
-- `name ` – microservice name
-- `version` – microservice version
-- `instanceId` – unique microservice instance id
+- `name ` – microservice name. All run instances of the same microservice must have the same name.
```
rkatsyuryna: running instances sounds better

doc/MSB.md
```diff
@@ -282,11 +282,13 @@ All configuration files use _key-value pair_ structure.
 ### Description of MSB configuration fields
 Service details section describes microservice parameters.
 
-- `name ` – microservice name
-- `version` – microservice version
-- `instanceId` – unique microservice instance id
+- `name ` – microservice name. All run instances of the same microservice must have the same name. 
+- `version` – microservice version. At the moment Msb-java doesn't handle the value.
```
rkatsyuryna: we use MSB-Java name not Msb-java

doc/MSB.md
rkatsyuryna: running

rkatsyuryna: i would write "different instanceId"  not "different instance ids"

doc/MSB.md
```diff
@@ -282,11 +282,13 @@ All configuration files use _key-value pair_ structure.
...
+- `instanceId` – unique microservice instance id. All run instances of the same microservice should have different instance ids.
 
-You may set these values directly
+Additional instances of a microservice can help with load balancing. The simplest way to have multiple instances is to deploy, configure and run a microservice from different locations with different instance ids. But if a microservice should be deployed and configured only once (for example, due to automatic load balancing reasons), particular instance id need to be absent in the config. The msb-java automatically generates unique instance id for each microservice run if the value is not present in the config.
```
rkatsyuryna: i would write "different instanceId" not "different instance ids",
"particular instanceId" not "particular instance id"
"unique instanceId" not "unique instance id" 
we use MSB-Java name not Msb-java

vso-tc: The statement about load balancing is a bit confusing because we can make different instances listen to a single queue not through **instanceId** but **groupId**.

PR #268, opened 2015-09-09
Removed redundant item in .gitignore

vso-tc:


.gitignore
```diff
@@ -15,5 +15,3 @@ logs/
 # Intellij #
 *.iml
 .idea/
-
-logs/
```
anstr, line 19: So from now logs directory could be committed into github?

vso-tc, line 19: Nope, if you look above, "logs" are already present in .gitignore file. This line was just a duplicate

PR #270, opened 2015-09-11
Added publishedAt field to the Message

anstr:


core/src/main/java/io/github/tcdl/msb/api/message/MetaMessage.java
```diff
@@ -40,8 +43,9 @@ public Builder(Integer ttl, Instant createdAt, ServiceDetails serviceDetails, Cl
         }     
 
         public MetaMessage build() {
-            Long durationMs = clock.instant().toEpochMilli() - this.createdAt.toEpochMilli();
-            return new MetaMessage(ttl, createdAt, durationMs, serviceDetails);
+            publishedAt = clock.instant();
+            Long durationMs = publishedAt.toEpochMilli() - this.createdAt.toEpochMilli();
```
vso-tc: We could get rid of subtraction by using [Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html) instead

core/src/test/java/io/github/tcdl/msb/message/MessageFactoryTest.java
```diff
@@ -219,4 +220,28 @@ public void testCreateAckBuilder() throws Exception {
...
+    @Test
+    public void testCreateRequestMessageBuilderPublishedAtPresent() {
+        String bodyText = "body text";
+        Payload requestPayload = TestUtils.createPayloadWithTextBody(bodyText);
+
+        Builder requestMessageBuilder = TestUtils.createMessageBuilder();
+
+        Message message = messageFactory.createRequestMessage(requestMessageBuilder, requestPayload);
+
+        assertNotNull(message.getMeta().getPublishedAt());
```
vso-tc, line 233: This test seems to be weaker than it could. Could we use fixed clock and check for exact value rather than verify nullability?

core/src/test/java/io/github/tcdl/msb/message/MessageFactoryTest.java
```diff
@@ -219,4 +220,28 @@ public void testCreateAckBuilder() throws Exception {
...
+    @Test
+    public void testCreateRequestMessageBuilderPublishedAtIsAfterCreatedAt() {
+        String bodyText = "body text";
+        Payload requestPayload = TestUtils.createPayloadWithTextBody(bodyText);
+
+        Builder requestMessageBuilder = TestUtils.createMessageBuilder();
+
+        Message message = messageFactory.createRequestMessage(requestMessageBuilder, requestPayload);
+
+        assertTrue(message.getMeta().getPublishedAt().isAfter(message.getMeta().getCreatedAt()));
```
vso-tc: Very long line. Let's split in using variables createdAt and publishedAt

vso-tc: Also I would suggest creating stronger test by using response messages when you created a responder and send 2 messages (ack and actual response). In such test you could check that that all 3 dates (createdAt and 2 publishedAt are one after another)

core/src/main/java/io/github/tcdl/msb/api/message/MetaMessage.java
anstr: Will be fixed.

core/src/test/java/io/github/tcdl/msb/message/MessageFactoryTest.java
anstr, line 233: We can't do this)Field "publishedAt" is calculated when method build() called and we don't know exact value.

core/src/test/java/io/github/tcdl/msb/message/MessageFactoryTest.java
anstr: > Very long line. Let's split in using variables createdAt and publishedAt

Will be fixed.

anstr: > Also I would suggest creating stronger test by using response messages when you created a responder and send 2 messages (ack and actual response). In such test you could check that that all 3 dates (createdAt and 2 publishedAt are one after another)

You mean IT test?

core/src/test/java/io/github/tcdl/msb/message/MessageFactoryTest.java
```diff
@@ -219,4 +220,32 @@ public void testCreateAckBuilder() throws Exception {
...
+        assertNotNull(message.getMeta().getPublishedAt());
+        assertEquals(message.getMeta().getPublishedAt(), FIXED_CLOCK.instant());
+    }
+
+    @Test
+    public void testCreateRequestMessageBuilderPublishedAtIsAfterCreatedAt() {
+        String bodyText = "body text";
+        Payload requestPayload = TestUtils.createPayloadWithTextBody(bodyText);
+
+        Builder requestMessageBuilder = TestUtils.createMessageBuilder(null);
```
vso-tc: I think it would be more expressive if you pass here `Clock.systemDefaultZone()` instead of `null`

core/src/test/java/io/github/tcdl/msb/support/TestUtils.java
```diff
@@ -180,12 +180,13 @@ public static Message createMsbResponseMessageWithAckNoPayload(Acknowledge ack,
                 .build();
     }
 
-    public static Message.Builder createMessageBuilder() {
+    public static Message.Builder createMessageBuilder(Clock clock) {
         MsbConfig msbConf = createMsbConfigurations();
-        Clock clock = Clock.systemDefaultZone();
+
+        Clock clock1 = (clock != null ? clock : Clock.systemDefaultZone());
```
vso-tc: You can get rid of this "if" if you implemented the suggestion above

core/src/test/java/io/github/tcdl/msb/message/MessageFactoryTest.java
```diff
@@ -219,4 +220,32 @@ public void testCreateAckBuilder() throws Exception {
...
+
+        Message message = messageFactory.createRequestMessage(requestMessageBuilder, requestPayload);
```
vso-tc, line 246: Unfortunately Java 8 doesn't have the suitable implementation of clock. Essentially what we need is a clock that returns an incremented value each time we call it. But we could implement it ourselves. Although I understand that it's far from scope of the current task.

For now please put a comment here why this test may blink.

anstr, line 246: ok

PR #273, opened 2015-09-11
minor: unit test for group id

rdrozdov-tc:


amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpAdapterFactory.java
```diff
@@ -59,7 +59,7 @@ protected AmqpBrokerConfig createAmqpBrokerConfig(MsbConfig msbConfig) {
                 .withFallback(amqpLibConfig);
 
         AmqpBrokerConfig brokerConfig = new AmqpBrokerConfig.AmqpBrokerConfigBuilder().withConfig(commonConfig).build();
-        if (brokerConfig != null && brokerConfig.getGroupId() == null) {
+        if (brokerConfig != null && (brokerConfig.getGroupId() == null || !brokerConfig.getGroupId().isPresent())) {
```
vso-tc: Let's simplify this condition

amqp/src/test/java/io/github/tcdl/msb/adapters/amqp/AmqpAdapterFactoryTest.java
```diff
@@ -137,6 +137,12 @@ public void testInit() {
     }
 
     @Test
+    public void testInitGroupIdWithServiceName() {
```
vso-tc: We should check positive case as well when groupId is present in the config

PR #283, opened 2015-10-27
Added logging

YuriySavchuk:


core/src/main/java/io/github/tcdl/msb/support/Utils.java
```diff
@@ -38,7 +38,9 @@ public static String validateTopic(String topic) {
         if (VALID_TOPIC_REGEXP.matcher(topic).matches()) {
             return topic;
         }
-        throw new IllegalArgumentException("\"" + topic + "\" must be an alpha-numeric, colon-delimited string");
+        String err = "Topic validation error: \"" + topic + "\" must be an alpha-numeric, colon-delimited string"; 
+        LOG.error(err);
```
vso-tc: Log and throw is considered to be anti-pattern: http://stackoverflow.com/questions/6639963/why-is-log-and-throw-condsidered-an-anti-pattern

PR #286, opened 2015-11-03
Fix for blinking test.

anstr:


core/src/test/java/io/github/tcdl/msb/support/TestClockWithStep.java
```diff
@@ -0,0 +1,76 @@
+package io.github.tcdl.msb.support;
+
+import java.io.Serializable;
+import java.time.Clock;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.temporal.TemporalUnit;
+
+public class TestClockWithStep extends Clock implements Serializable {
```
vso-tc: What do you think about more generic name like `IncrementingClock`?

anstr: Agree.

PR #295, opened 2015-11-24
Requester should be able to wait for acks even if configured to waitF…

rkatsyuryna:
also fixes from msb pull #55 are added


core/src/main/java/io/github/tcdl/msb/collector/Collector.java
```diff
@@ -42,7 +41,8 @@
 
     private int timeoutMs;
     private int currentTimeoutMs;
-    private long waitForAcksUntil;
+    private Integer waitForAcksMs;
+    private Long waitForAcksUntil;
```
wheleph: Essentially this field represents some moment of time. Maybe it's more natural and expressive to use `java.time.Instant` instead of plain `java.lang.Long`?

core/src/main/java/io/github/tcdl/msb/collector/Collector.java
```diff
@@ -42,7 +41,8 @@
...
     private int waitForResponses;
```
wheleph: we can easily get rid of field `waitForResponses`. It's used nowhere except of the constructor

core/src/main/java/io/github/tcdl/msb/collector/Collector.java
rkatsyuryna: agree)) will fix this.Fixed

core/src/main/java/io/github/tcdl/msb/collector/Collector.java
rkatsyuryna: yes. Fixed

core/src/main/java/io/github/tcdl/msb/collector/Collector.java
```diff
@@ -253,6 +238,30 @@ private int setResponsesRemainingForResponderId(String responderId, Integer resp
         return responsesRemainingById.get(responderId);
     }
 
+    private int getWaitForResponsesFromConfigs(RequestOptions requestOptions) {
+        if (requestOptions.getWaitForResponses() == null || requestOptions.getWaitForResponses() == WAIT_FOR_RESPONSES_UNTIL_TIMEOUT) {
+            // use for Infinity number or expected responses
+            return Integer.MAX_VALUE;
```
wheleph: Is `MAX_VALUE` large enougth? Looks like this logic might be broken if we get back more than 2147483647 responses. Why bother with that? You have a whole spectrum of negative numbers to describe this semantic (waiting for responses until timeout reached).

core/src/main/java/io/github/tcdl/msb/impl/RequesterImpl.java
```diff
@@ -109,6 +109,14 @@ public void publish(Object requestPayload, Message originalMessage, String tag)
         }
     }
 
+    private boolean isWaitForAckMs() {
+        return requestOptions.getAckTimeout() != null && requestOptions.getAckTimeout() != 0;
```
wheleph, line 113: It looks tricky. You work with fields of `RequestOptions` in a non-trivial way here and in `getWaitForResponsesFromConfigs`. The non-trivial semantics of values (if there's any) should be gathered in a single place.

core/src/main/java/io/github/tcdl/msb/collector/Collector.java
rkatsyuryna: ok in Requester from Node.js we use  if (this.waitForAcksMs || this.waitForResponses) ...in java version we read requestOption parameters..will do some refactoring to have both values in sync in both classes

core/src/test/java/io/github/tcdl/msb/impl/RequesterImplTest.java
```diff
@@ -211,12 +218,12 @@ public void testRequestMessageWithTags() throws Exception {
...
     }
 
-    private RequesterImpl<RestPayload> initRequesterForResponsesWithTimeout(int numberOfResponses) throws Exception {
+    private RequesterImpl<RestPayload> initRequesterForResponsesWith(Integer numberOfResponses, Integer respTimeout,  Integer ackTimeout , Callback<Void> endHandler) throws Exception {
 
         MessageTemplate messageTemplateMock = mock(MessageTemplate.class);
 
         RequestOptions requestOptionsMock = new RequestOptions.Builder().withMessageTemplate(messageTemplateMock).withWaitForResponses(numberOfResponses)
-                .withResponseTimeout(100).build();
+                .withResponseTimeout(respTimeout).withAckTimeout(ackTimeout).build();
```
wheleph: It would be better to split "with" and "build" method invokations into separate lines.

core/src/main/java/io/github/tcdl/msb/collector/Collector.java
rkatsyuryna: we have increment logic that does waitForResponses -1 for each incomming message
2 billion is good enough to be used

core/src/test/java/io/github/tcdl/msb/impl/RequesterImplTest.java
```diff
@@ -81,9 +77,20 @@ public void testPublishWaitForResponses() throws Exception {
     }
 
     @Test
+    public void testPublishWaitForResponsesAck() throws Exception {
+        RequesterImpl<RestPayload> requester = initRequesterForResponsesWith(1, 1000, 800, arg ->  fail());
```
wheleph, line 81: It seems like this unit test doesn't cover the original requirement: wait for acks even if waitForResponses is 0. Can we have such test here?

core/src/main/java/io/github/tcdl/msb/collector/Collector.java
wheleph: Still I don't understand. What is going to happen if you set waitForRespones -1, but you get back 3 billion respones? Will you be able to process them all?

rkatsyuryna: OK. Fixed

core/src/main/java/io/github/tcdl/msb/impl/RequesterImpl.java
rkatsyuryna, line 113: changed to much cleaner aproach

core/src/test/java/io/github/tcdl/msb/impl/RequesterImplTest.java
rkatsyuryna: ok.Fixed

core/src/test/java/io/github/tcdl/msb/impl/RequesterImplTest.java
rkatsyuryna, line 81: looking..

core/src/main/java/io/github/tcdl/msb/impl/RequesterImpl.java
rkatsyuryna, line 113: fixed

core/src/test/java/io/github/tcdl/msb/impl/RequesterImplTest.java
rkatsyuryna, line 81: it does we create collector with expectation of one message, then we received this message (handleMessage() ) and verify end() event was not fired (arg ->  fail())

vso-tc, line 81: well, yeah. But this ticket is about fix when waitForResponses is 0, not 1. I suspect that the test with 1 would pass even without the fix implemented.

rkatsyuryna, line 81: yes but we two have tests that cove this. This one and testHandleResponseNoResponsesRemainingAndWaitUntilAckBeforeNow()

rkatsyuryna, line 81: one is testing the initial configuration waitForResponses **is** 0 and we waitForAcksTimeout and this one that we had dynamicly changed status(last message was received) from 1 to 0 and waitForResponses **become** 0:smile:

core/src/main/java/io/github/tcdl/msb/collector/Collector.java
```diff
@@ -253,6 +242,22 @@ private int setResponsesRemainingForResponderId(String responderId, Integer resp
...
+        int newTimeoutMs = this.currentTimeoutMs - toIntExact(clock.instant().toEpochMilli() - this.startedAt);
+        LOG.debug("Waiting for responses until {}.", clock.instant().plus(newTimeoutMs, ChronoUnit.MILLIS));
+        this.responseTimeoutFuture = timeoutManager.enableResponseTimeout(newTimeoutMs, this);
+    }
+
+    void waitForAcks() {
+        if (ackTimeoutFuture == null) {
+            LOG.debug("Waiting for ack until {}.", this.waitForAcksUntil);
+            long ackTimeoutMs = waitForAcksUntil.toEpochMilli() - clock.instant().toEpochMilli();
+            ackTimeoutFuture = timeoutManager.enableAckTimeout(toIntExact(ackTimeoutMs), this);
```
YuriySavchuk, line 255: Probably, we need to check waitForAcksUntil is not null again for prevention NPE (for example in tests).

rkatsyuryna, line 255: really don't got u poin. We call isAwaitingAcks() before call to this method that contains check for null. In case null we never call this method. This method is package-private for testing puropse and why do we ever need null check for testing?? Please provide more details

PR #296, opened 2015-12-04
Amqp confirms rejects

YuriySavchuk:


amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpAcknowledgementHandler.java
```diff
@@ -0,0 +1,81 @@
...
+            LOG.warn(ACK_WAS_ALREADY_SENT);
+        }
+    }
+
+    @Override
+    public void rejectMessage() {
+        if (!isAcknowledgementSent) {
+            try {
+                channel.basicReject(deliveryTag, isRequeueRejectedMessages);
+                LOG.error(String.format("[consumer tag: %s] AMQP reject has been sent for message: %s", consumerTag, bodyStr));
```
anha1: Please consider using "debug" as far as message reject is a valid option of a client.

core/src/main/java/io/github/tcdl/msb/api/Responder.java
```diff
@@ -1,11 +1,12 @@
 package io.github.tcdl.msb.api;
 
+import io.github.tcdl.msb.adapters.ConsumerAdapter;
 import io.github.tcdl.msb.api.message.Message;
 
 /**
  * Responsible for creating responses and acknowledgements and sending them to the bus.
  */
-public interface Responder {
+public interface Responder extends ConsumerAdapter.AcknowledgementHandler {
```
anha1: Please consider using ackHandler getter instead so a client could use it directly.

amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpAcknowledgementHandler.java
```diff
@@ -0,0 +1,77 @@
...
+        super();
+        this.channel = channel;
+        this.consumerTag = consumerTag;
+        this.deliveryTag = deliveryTag;
+        this.isRequeueRejectedMessages = isRequeueRejectedMessages;
+    }
+
+    @Override
+    public void confirmMessage() {
+        if (!isAcknowledgementSent) {
```
wheleph: This method is not thread safe which may lead to problems with concurrent acknowldegements. If this class is meant to be non-thread-safe it must be specified in javadoc for `AcknowledgementHandler` because it's part of client-facing API.

amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpAcknowledgementHandler.java
```diff
@@ -0,0 +1,77 @@
...
+                isAcknowledgementSent = true;
+            } catch (Exception e) {
+                LOG.error(String.format("[consumer tag: %s] Got exception when trying to reject a message:", consumerTag), e);
+            }
+        } else {
+            LOG.warn(ACK_WAS_ALREADY_SENT);
+        }
+    }
+
+    public void autoConfirm() {
```
wheleph, line 91: Why do we need these non-interface methods? Could we just invoke `confirmMessage` and `autoReject` from whatever place?

amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpMessageProcessingTask.java
```diff
@@ -34,10 +33,13 @@ public AmqpMessageProcessingTask(String consumerTag, String body, Channel channe
     public void run() {
         try {
             LOG.debug(String.format("[consumer tag: %s] Starting message processing: %s", consumerTag, body));
-            msgHandler.onMessage(body);
+            msgHandler.onMessage(body, ackHandler);
             LOG.debug(String.format("[consumer tag: %s] Message has been processed: %s", consumerTag, body));
+            ackHandler.autoConfirm();
```
wheleph, line 38: I would add logging after invocation of `autoConfirm` and `autoReject` to have evidence in logs that those calls completed succesfully

amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpMessageConsumer.java
```diff
@@ -41,27 +43,34 @@ public AmqpMessageConsumer(Channel channel, ExecutorService consumerThreadPool,
...
             LOG.debug(String.format("[consumer tag: %s] Message consumed from broker: %s", consumerTag, bodyStr));
+
             try {
-                consumerThreadPool.submit(new AmqpMessageProcessingTask(consumerTag, bodyStr, getChannel(), envelope.getDeliveryTag(), msgHandler));
-                LOG.debug(String.format("[consumer tag: %s] Message has been put in the processing queue: %s. About to send AMQP ack...",
+                consumerThreadPool.submit(new AmqpMessageProcessingTask(consumerTag, bodyStr, msgHandler, ackHandler));
+                LOG.debug(String.format("[consumer tag: %s] Message has been put in the processing queue: %s.",
                         consumerTag, bodyStr));
-
-                getChannel().basicAck(envelope.getDeliveryTag(), false);
```
wheleph, line 54: It seems like acknowledgement logic is changed here and we auto-ack messages not when they are consumed from queue but when the task from thread pool gets executed. I think it is really cool but we probably need to change the logic for the shutdown of the thread pool. We no longer need to wait until all tasks from the queue complete. It's sufficient to wait for the completion of only the running tasks.

amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpMessageProcessingTask.java
wheleph, line 38: **MAJOR** So we invoke `autoConfirm` (and hence fire AMQP ack) unconditionally after `onMessage` exit. Is it correct? What if the developer wants to fire an explicit ack in a separate thread, will it mean that the ack will be sent prematurely? Am I missing something here?

core/src/main/java/io/github/tcdl/msb/Consumer.java
```diff
@@ -96,12 +99,14 @@ protected void handleRawMessage(String jsonMessage) {
             LOG.debug("Message has been successfully parsed {}", jsonMessage);
 
             if (!isMessageExpired(message)) {
-                messageHandler.handleMessage(message);
+                messageHandler.handleMessage(message, acknowledgeHandler);
             } else {
                 LOG.warn("Expired message: {}", jsonMessage);
+                acknowledgeHandler.rejectMessage();
```
wheleph, line 105: **MAJOR** Depending on settings the handler may reject and requeue the message which should not be the case here. IMHO this handler should always discard messages.

core/src/main/java/io/github/tcdl/msb/api/ResponderContext.java
```diff
@@ -0,0 +1,25 @@
+package io.github.tcdl.msb.api;
+
+import io.github.tcdl.msb.api.message.Message;
+
+/**
+ * Provides access to Responder Context.
+ */
+public interface ResponderContext {
```
wheleph: I don't get why we introduced `ResponderContext`. It seems to be not symmetrical to what we have for `Requester` where we pass ack handler explicitly.

amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpMessageConsumer.java
```diff
@@ -41,27 +43,34 @@ public AmqpMessageConsumer(Channel channel, ExecutorService consumerThreadPool,
...
         } catch (Exception e) {
             // Catch all exceptions to prevent AMQP channel to be closed
-            LOG.error(String.format("[consumer tag: %s] Got exception while processing incoming message", consumerTag), e);
+            LOG.error(String.format("[consumer tag: %s] Got exception while processing incoming message. About to send AMQP reject...", consumerTag), e);
+            ackHandler.autoReject();
         }
     }
+    
+    AmqpAcknowledgementHandler createAcknowledgementHandler(Channel channel, String consumerTag, long deliveryTag, boolean isRequeueRejectedMessages) {
+        return new AmqpAcknowledgementHandler(channel, consumerTag, deliveryTag, isRequeueRejectedMessages);
```
wheleph, line 73: **MAJOR** We might have a problem here. Because we create handler per consumer. Suppose `isRequeueRejectedMessages` is  `true`. But in some cases (for example when the message cannot be parsed) it should not be requeued so we should use some other hander for that consumer. I see 2 ways to overcome this limitation:
- introduce another method in the handler interface `dropMessage`
- invoke `confirmMessage` in the cases described above

cli/src/main/resources/application.conf
```diff
@@ -22,6 +22,7 @@ msbConfig {
     # -1 means unlimited
     consumerThreadPoolQueueCapacity = 20
     requeueRejectedMessages = true
+    prefetchCount = 0
```
wheleph, line 25: **MAJOR** Why didn't we introduce another key to switch between automatic and explicit acknowledgement? It seems like we need one.

wheleph:
It seems like we do need to update the devguide because:
- we have quite complex logic for automatic rejects. It should be understandable by the users of the library
- due to the fact that we have single set of options for all consumers in a microservice (limiation of the library) there're some non-trivial use cases. For example let's say that we have a microservice with responder server that users requester inside of it. We want to make the responder server use explicit acks. It means that the developer also must use explicit acks when processing responses for the requester. Otherwise those responses will be left in queues forever.

amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpAcknowledgementHandler.java
```diff
@@ -0,0 +1,77 @@
...
+    public void confirmMessage() {
+        if (!isAcknowledgementSent) {
+            try {
+                channel.basicAck(deliveryTag, false);
+                isAcknowledgementSent = true;
+            } catch (Exception e) {
+                LOG.error(String.format("[consumer tag: %s] Got exception when trying to confirm a message:", consumerTag), e);
+            }
+        } else {
+            LOG.warn(ACK_WAS_ALREADY_SENT);
```
wheleph: I agree that just logging warning for the duplicate confirm is ok. But what if developer invoked `confirm` and then `reject`? I think it clearly indicates some problem in microservice's logic and we should have some more evidence in logs then just a WARN message

amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpAcknowledgementHandler.java
YuriySavchuk: Refactored. Used AtomicBoolean for unblocked synchronization

amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpAcknowledgementHandler.java
YuriySavchuk, line 91: confirmMessage() logs ACK_WAS_ALREADY_SENT. In case where autoConfirm() is used it's not necessary.

amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpMessageProcessingTask.java
YuriySavchuk, line 38: Good point. A solution is to use setAutoAcknowledgement(false) if message processing need to be continued out of current thread. MS developer is responsible for explicit confirm/reject in this case. This point will be described in the Dev guide.

amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpMessageConsumer.java
YuriySavchuk, line 73: Development was continued and solution with "Reject message without requeue" was implemented.

cli/src/main/resources/application.conf
YuriySavchuk, line 25: prefetchCount - It's not a key. We don't use AutoAcknowledgement configuration option. Only explicit AmqpAcknowledgementHandler.setAutoAcknowledgement(boolean autoAcknowledgement) method is used for that.

core/src/main/java/io/github/tcdl/msb/Consumer.java
YuriySavchuk, line 105: Expired messages are discarded

core/src/main/java/io/github/tcdl/msb/api/ResponderContext.java
YuriySavchuk: See "Added MessageContext" commit

amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpAcknowledgementHandler.java
YuriySavchuk: Now ACK_WAS_ALREADY_SENT is logged with error level

YuriySavchuk: Dev guide update is in the ToDo list. Will try to update it ASAP

PR #300, opened 2015-12-11
Described 1.4.x changes

YuriySavchuk:


release-notes.html
```diff
@@ -1,19 +1,23 @@
 <html>
 <head>
   <meta http-equiv="content-type" content="">
-  <title>Welcome to MSB-Java version 1.3.0</title>
```
wheleph, line 4: Why HTML? Maybe md would be better?

PR #304, opened 2015-12-16
WEB-16678 time-consuming response callbacks no longer block incoming …

anha1:
WEB-16678 time-consuming response callbacks no longer block incoming responses from being processed

### DIFF
--- a/amqp/pom.xml
+++ b/amqp/pom.xml
@@ -27,6 +27,11 @@
             <artifactId>msb-java-core</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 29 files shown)


amqp/pom.xml
```diff
@@ -27,6 +27,11 @@
             <artifactId>msb-java-core</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.github.tcdl.msb</groupId>
+            <artifactId>msb-java-core</artifactId>
```
rkatsyuryna, line 31: why do we need this dependency in amqp?

PR #315, opened 2015-12-25
Fix:channel monitor agent throws exceptions upon microservices shutdown

rkatsyuryna:


rkatsyuryna:
Making exchange auto-delete=false will prevent this kind of exceptions in future: When we have single consumer service shutdown/in restart(e.g redeployment) will not cause producer service to throw exception

PR #317, opened 2016-01-04
Use of logback in microservices

rkatsyuryna:


anha1:
These changes will be merged later using a different branch.

PR #331, opened 2016-02-08
forward namespace support

anha1:


core/src/main/java/io/github/tcdl/msb/api/message/Topics.java
```diff
@@ -8,12 +8,14 @@
 
     private final String to;
     private final String response;
+    private final String forward;
 
     @JsonCreator
-    public Topics(@JsonProperty("to") String to, @JsonProperty("response") String response) {
+    public Topics(@JsonProperty("to") String to, @JsonProperty("response") String response,  @JsonProperty("forward") String forward) {
```
alexandr-zolotov, line 14: I'd prefer to static factory method annotated JsonCreator and chained properties (like new Topics().to("proxy").forward("final_destination") )

PR #340, opened 2016-03-01
performance improvements

anha1:


amqp/src/main/resources/amqp.conf
```diff
@@ -20,7 +20,7 @@ config.amqp = {
   consumerThreadPoolQueueCapacity = -1
 
   # Interval of the heartbeats that are used to detect broken connections. Zero for none. See for more details: https://www.rabbitmq.com/heartbeats.html
-  heartbeatIntervalSec = 1
+  heartbeatIntervalSec = 30
```
anha1, line 23: with this 1s delay msb-java loses connection under a heavy load

core/src/main/java/io/github/tcdl/msb/collector/CollectorManager.java
```diff
@@ -53,26 +53,21 @@ public void registerCollector(Collector collector) {
...
      */
     public void unregisterCollector(Collector collector) {
         collectorsByCorrelationId.remove(collector.getRequestMessage().getCorrelationId());
-
-        synchronized (this) {
-            if (collectorsByCorrelationId.isEmpty() && isSubscribed) {
-                channelManager.unsubscribe(topic);
-                isSubscribed = false;
-            }
-        }
```
anha1, line 75: As far as responses topic is static, I don't see a need in this "unsubscripe" call that has a significant negative performance impact (a lot of subscribe/unsubscribe cycles). I'll create msb-node change request regarding it.

amqp/src/main/resources/amqp.conf
alexandr-zolotov, line 23: Don't you know why?

anha1, line 23: The application can't deliver heartbeats each second under a heavy load, so RabbitMQ closes a connection on a missing heartbeat. AFAIR, we faced this issue before during our MS load testing. So I'm considering the previous 1s default value as dangerous and harmful. RabbitMQ defaults are also far from 1s: "The timeout is in seconds, and default value is 60 (580 prior to release 3.5.5).".

PR #344, opened 2016-03-18
Web 19507

YuriySavchuk:


core/src/main/java/io/github/tcdl/msb/config/ServiceDetails.java
```diff
@@ -67,19 +69,27 @@ public ServiceDetails build() {
             return new ServiceDetails(name, version, instanceId, hostname, ip, pid);
         }
 
-        private static InetAddress getHostInfo() {
-            InetAddress hostInfo = null;
+        protected InetAddress getHostInfo() {
```
alexandr-zolotov, line 72: I suggest to use Optional<InetAddress> as return type because it clearly shows that calling code may not get what it expects. Than the code will look as follows

```
try {
    return Optional.of(InetAddress.getLocalHost());
} catch (UnknownHostException ex) {
    LOG.warn("Fail to retrieve host info", ex);
    return Optional.empty();
}
```

PR #364, opened 2016-10-07
AMQP explicit exchange type support

alexandr-zolotov:


sergiv83:
Good work! I see all the things we've discussed implemented. I mean the ability to set exchange type explicitly, ResponderOption object and it's AMQP child, more convenient method signatures and depreciation of old signatures. There is only one thing that would be nice to have from my point of view is updated examples. Because it is an entry point and kind of reference for start working with the library, but approaches shown there now are outdated. That is a bit confusing.

shumsky (approved):

alexandr-zolotov:
@sergiv83 , could you please point a bit more precisely which part of examples is outdated? I fail to figure it out.

sergiv83:
@alexandr-zolotov , DateExtractor, FacetsAggregator and the most important PongService :) are using deprecated createResponderServer(...) method

examples/src/main/java/io/github/tcdl/msb/examples/PongService.java
```diff
@@ -14,7 +14,8 @@ public static void main(String[] args) {
 
         ObjectFactory objectFactory = msbContext.getObjectFactory();
         MessageTemplate messageTemplate = new MessageTemplate().withTags("pong-static-tag");
-        ResponderServer responderServer = objectFactory.createResponderServer("pingpong:namespace", messageTemplate,
+        ResponderOptions responderOptions = new ResponderOptions.Builder().withMessageTemplate(messageTemplate).build();
+        ResponderServer responderServer = objectFactory.createResponderServer("pingpong:namespace", responderOptions,
                 (request, responderContext) -> {
```
agritsik, line 19: I believe, we should simplify it. Get rid of boilerplates such as ObjectFactory, MessageTemplate, ResponderOptions from the client's code etc

Finally, I'd like to get something like that:

```
msbBuilder.createRequester().publish("hello!");
msbBuilder.createResponder(r->print( r ));
```

Methods createResponder and createRequester should have an ability to get ResponseOptions/RequestOptions object as an optional parameter though. 

If you agree, we can create a separate task for this. Plus we need to align the solution with node.js version

PR #367, opened 2016-11-01
Autoconfigure module for spring boot

sergiv83:


spring-boot-starter/src/main/java/io/github/tcdl/msb/autoconfigure/MsbConfigAutoConfiguration.java
```diff
@@ -0,0 +1,108 @@
...
+    public MsbConfig config() {
+        Config config = ConfigFactory.load("reference");
+        String appName = StringUtils.isNotBlank(msbProperties.serviceDetails.name) ? msbProperties.serviceDetails.name : DEFAULT_APP_NAME;
+        config = config.withValue("msbConfig.serviceDetails.name", ConfigValueFactory.fromAnyRef(appName));
+        config = config.withValue("msbConfig.serviceDetails.instanceId", ConfigValueFactory.fromAnyRef(msbProperties.serviceDetails.instanceId));
+
+        // Service Details
+        String version = StringUtils.isNotBlank(msbProperties.serviceDetails.version) ? msbProperties.serviceDetails.version : DEFAULT_VERSION;
+        config = config.withValue("msbConfig.serviceDetails.version", ConfigValueFactory.fromAnyRef(version));
+        if (StringUtils.isNoneBlank(msbProperties.serviceDetails.hostname))
```
alexandr-zolotov: I think you meant isNotBlank here.

alexandr-zolotov: In general, I don't fully understand what is the benefit of using Spring type checked config if you put it as a string in TypeSafe config afterwards. To me it looks like *.yml / *.properties file + general mechanism for overriding would do the job. Plus, Spring supports default values for such files. You wouldn't need all this isNotBlank and null checks.

sergiv83: 1) yes, it was a typo
2) I don't want to have more than one properties/yml file for one configuration to keep things simple, and I don't want to force msb-client to have a configuration at all

PR #392, opened 2017-10-24
IN-3160: exposing the message count in the Consumer

KimEeckhout:


amqp/src/test/java/io/github/tcdl/msb/adapters/amqp/AmqpConsumerAdapterTest.java
```diff
@@ -227,6 +227,64 @@ public void testIsDurableTrueIfNotResponseTopicAndDurableConfig() throws IOExcep
...
+
+        adapter.subscribe((jsonMessage, ackHandler) -> {
+        });
+
+        verify(mockChannel).exchangeDeclare(topicName, "fanout", false, true, null);
+
+        Optional<Long> answer = Optional.of(42L);
+        Optional<Long> result = adapter.messageCount();
+        verify(mockChannel, times(1)).messageCount(anyString());
+        assertEquals(result, answer);
```
b-dzoba: nit: Expected value should be the first argument in assertEquals.

amqp/src/main/java/io/github/tcdl/msb/adapters/amqp/AmqpConsumerAdapter.java
```diff
@@ -20,10 +21,10 @@
     private String consumerTag;
     private AmqpBrokerConfig adapterConfig;
     private boolean isResponseTopic = false;
+    private Optional<String> currentQueueName = Optional.empty();
```
b-dzoba, line 24: I would not use Optional for fields (as a return value Optional is fine).

```
Optional is primarily intended for use as a method return type where there is a clear need to represent "no result," and where using null is likely to cause errors.
```

KimEeckhout: What would you like to see instead? I can't think of a better way to express a field value that may or may not be present depending on the state of the instance.

b-dzoba: I would just use String for field type. In this case absence of value is represented by null.

amqp/src/test/java/io/github/tcdl/msb/adapters/amqp/AmqpConsumerAdapterTest.java
KimEeckhout: Indeed. Fixed now.

b-dzoba: @Mopper, Please fix argument order in other assert statements as well.

b-dzoba:
@Mopper, Can you explain how this functionality can be used? As I understand, Consumer is an internal class and users of MSB API don't have access to it.

rdrozdov-tc:
How do you propose to call this message count? Currently customer is not exposed to be used from msb API

KimEeckhout:
Good point @b-dzoba and @rdrozdov-tc . I somehow got it in my head that we use the Consumer directly but that's clearly wrong. Back to the drawing board!

PR #399, opened 2018-09-06
Disable MSB-Java using property

jabengozar:
Disable MSB-Java using property

- true or not present means enable (back compatibility)
- explicit false means disable
- Updated documentation
- Added test

### DIFF
--- a/doc/MSB.md
+++ b/doc/MSB.md
@@ -285,6 +285,10 @@ All configuration files use _key-value pair_ structure.
 - [application.conf](/acceptance/src/main/resources/application.conf) - overrides values from reference.conf
 
 ### Description of MSB configuration fields
+
+There is a feature flag for enable/disable MSB:
+- `msb-config.enabled` true or missing value means enabled. false for disable whole MSB. It's very useful for testing and for enable/disable the msb capabilities depending of environments (f.e. as feature flag if the functionality is not desirable and/or not tested yet).
+
 Service details section describes microservice parameters.
 
 - `name ` – microservice name. All running instances of the same microservice must have the same name. Mandatory, has no default value.
--- a/spring-boot-starter/src/main/java/io/github/tcdl/msb/autoconfigure/MsbConfigAutoConfiguration.java
+++ b/spring-boot-starter/src/main/java/io/github/tcdl/msb/autoconfigure/MsbConfigAutoConfiguration.java
@@ -8,10 +8,12 @@ import io.github.tcdl.msb.support.Utils;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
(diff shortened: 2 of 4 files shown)


sergiv83 (changes requested):
@jabengozartc 
Do I understand right that the reason of this PR is to add an ability to disable autoconfiguration through the specific property value (msb-config.enabled)? If yes, than maybe it's easier to just exclude msb-spring-boot-starter from your pom.xml?
Moreover, I have 2 less critical points:
* it would be good to cover the new functionality with test (inside spring-boot-starter module)
* I think [MSB.md](https://github.com/tcdl/msb-java/blob/master/doc/MSB.md) should be updated to describe the impact of msb-config.enabled

jabengozar:
Yes, That's the reason, but we want enable/disable with properties not in `pom.xml`, because we want this mechanism depending on the environment.
Also prevent create a mock in test adding `msb-config.enabled=false` in `src/test/resources/application.properties`
By the way in our project that use this one, we disable the msbConfig in test in this way:
```
@Configuration
public class MsbConfiguration {

    @MockBean
    public MsbContext msbContext;

}
```
`@MockBean` create a mock and put in automatically in the context replacing the existing one :)

I'm going to review the markdown file, create a test to cover and update this PR

Thanks for the feedback and the work in this project.

sergiv83 (approved):

b-dzoba:
You should be able to use property `spring.autoconfigure.exclude` to exclude auto configuration:
```
spring.autoconfigure.exclude=io.github.tcdl.msb.autoconfigure.MsbConfigAutoConfiguration
```
There is no need for additional property.

jabengozar:
> You should be able to use property spring.autoconfigure.exclude to exclude auto configuration:
> 
> spring.autoconfigure.exclude=io.github.tcdl.msb.autoconfigure.MsbConfigAutoConfiguration
> There is no need for additional property.

Yes, you are right, but it's a general purpose property, I mean, we can use it for prevent loading of whatever. Also need to pass the package.

I'd like have something more particular, like feature flag, that could be resolved using a env var, for example:
`msb-config.enabled=${MSB_CONFIG_ENABLED:false}`
and set the value per environment in AWS Parameter store, for example.
In the common property (Spring documentation) there are a lot of feature flags (ends in `.enabled`) that activate the module which belonging to:
https://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html

What's your opinion?
Thanks

b-dzoba:
I am not convinced that we should add another property to disable auto-configuration as there are other ways to achieve the same thing in Spring Boot (https://docs.spring.io/spring-boot/docs/current/reference/html/using-boot-auto-configuration.html#using-boot-disabling-specific-auto-configuration).

> In the common property (Spring documentation) there are a lot of feature flags (ends in .enabled) that activate the module which belonging to:

I think this is mostly used to configure features without disabling them completely. Spring's own auto-configurations usually don't have enable property (e.g. 
https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.html). 

> msb-config.enabled

Another point, is that this property can be confusing because it will only disable auto-configuration for MSB, not user's own configuration.

> I'd like have something more particular, like feature flag, that could be resolved using a env var
and set the value per environment in AWS Parameter store, for example.

Not sure why you would want to do this, but you can always create separate configuration in the application. E.g.:
```
@Configuration
@ConditionalOnProperty(value = "msb-config.enabled", havingValue = "false")
@EnableAutoConfiguration(exclude={MsbConfigAutoConfiguration.class})
public class MsbDisabledConfiguration {
}
```

jabengozar:
thanks!!

